PR #2, opened 2019-04-02
Implement CI scripts for TPM 1.2 simulator and tcsd.

twitchy-jsonp:


ci/setup_tpm12_simulator.sh
```diff
@@ -63,8 +63,46 @@ fetch_simulator () {
...
+  echo "${SIM_PID}" > "${BUILD_BASE}/sim_pid"
+  disown
+  sleep 2
+}
+
+setup_tpm () {
+  echo "Starting the TPM..."
+  ${SIMULATOR_SRC}/libtpm/utils/tpmbios -v
+  echo "Creating a fake EK..."
+  ${SIMULATOR_SRC}/libtpm/utils/createek || true
```
mjg59: Why || true? If this fails we presumably want to bail

ci/setup_tpm12_simulator.sh
```diff
@@ -63,8 +63,46 @@ fetch_simulator () {
...
+  cd "${SIMULATOR_SRC}/libtpm" && make -j4
+}
+
+run_simulator () {
+  mkdir -pv "${BUILD_BASE}/NVRAM"
+  export TPM_PORT='6545'
+  export TPM_PATH="${BUILD_BASE}/NVRAM"
+  export TPM_SERVER_NAME='localhost'
+  export TPM_SERVER_PORT='6545'
+  export TCSD_TCP_DEVICE_PORT='6545'
```
mjg59: Should this be in run_tcsd? Possibly define 6545 as a constant elsewhere.

ci/setup_tpm12_simulator.sh
twitchy-jsonp: This always fails with `Error TPM deactivated from TPM_CreateEndorsementKeyPair` - I think this is expected. I've got this working in redshell, I'll know for sure once its working in Kokoro.

ci/setup_tpm12_simulator.sh
twitchy-jsonp: Done.

PR #10, opened 2019-04-08
Aik hierarchy

DenisKarch:
Move AIK to Storage hierarchy.

### DIFF
--- a/attest/tpm_linux.go
+++ b/attest/tpm_linux.go
@@ -190,15 +190,21 @@ func (t *TPM) getPrimaryKeyHandle(pHnd tpmutil.Handle) (tpmutil.Handle, bool, er
 		return pHnd, false, nil
 	}
 
-	ekHnd, _, err := tpm2.CreatePrimary(t.rwc, tpm2.HandleEndorsement, tpm2.PCRSelection{}, "", "", defaultEKTemplate)
-	if err != nil {
-		return 0, false, fmt.Errorf("EK CreatePrimary failed: %v", err)
+	var keyHnd tpmutil.Handle
+	switch pHnd {
+	case commonSrkEquivalentHandle:
+		keyHnd, _, err = tpm2.CreatePrimary(t.rwc, tpm2.HandleOwner, tpm2.PCRSelection{}, "", "", defaultSRKTemplate)
+	case commonEkEquivalentHandle:
+		keyHnd, _, err = tpm2.CreatePrimary(t.rwc, tpm2.HandleEndorsement, tpm2.PCRSelection{}, "", "", defaultEKTemplate)
 	}
-	defer tpm2.FlushContext(t.rwc, ekHnd)
-
-	err = tpm2.EvictControl(t.rwc, "", tpm2.HandleOwner, ekHnd, pHnd)
 	if err != nil {
-		return 0, false, fmt.Errorf("EK EvictControl failed: %v", err)
+		return 0, false, fmt.Errorf("CreatePrimary failed: %v", err)
+	}
+	defer tpm2.FlushContext(t.rwc, keyHnd)
+
+	err = tpm2.EvictControl(t.rwc, "", tpm2.HandleOwner, keyHnd, pHnd)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


attest/tpm_linux.go
```diff
@@ -190,7 +190,13 @@ func (t *TPM) getPrimaryKeyHandle(pHnd tpmutil.Handle) (tpmutil.Handle, bool, er
 		return pHnd, false, nil
 	}
 
-	ekHnd, _, err := tpm2.CreatePrimary(t.rwc, tpm2.HandleEndorsement, tpm2.PCRSelection{}, "", "", defaultEKTemplate)
+	var ekHnd tpmutil.Handle
+	switch pHnd {
+	case commonSrkEquivalentHandle:
+		ekHnd, _, err = tpm2.CreatePrimary(t.rwc, tpm2.HandleOwner, tpm2.PCRSelection{}, "", "", defaultSRKTemplate)
```
twitchy-jsonp: Please revise the name to be more generic.

attest/tpm_linux.go
```diff
@@ -393,7 +407,7 @@ func (t *TPM) MintAIK(opts *MintOptions) (*Key, error) {
 		}()
 
 		// We can only certify the creation immediately afterwards, so we cache the result.
-		attestation, sig, err := tpm2.CertifyCreation(t.rwc, "", keyHandle, keyHandle, nil, creationHash, tpm2.SigScheme{tpm2.AlgRSASSA, tpm2.AlgSHA256, 0}, tix)
+		attestation, sig, err := tpm2.CertifyCreation(t.rwc, "", keyHandle, keyHandle, nil, creationHash, tpm2.SigScheme{tpm2.AlgRSASSA, tpm2.AlgSHA256, 0}, &tix)
```
twitchy-jsonp: Does it work without taking the pointer to `tix`?

attest/tpm_linux.go
```diff
@@ -408,6 +422,7 @@ func (t *TPM) MintAIK(opts *MintOptions) (*Key, error) {
 			KeyEncoding:       KeyEncodingParameterized,
```
twitchy-jsonp: Should this be a different `KeyEncoding`?

twitchy-jsonp: (such as `KeyEncodingEncrypted`)

attest/tpm_linux.go
twitchy-jsonp: (`ekHnd`)

attest/tpm_linux.go
DenisKarch: nope, would not build with just tix. (CreateKeyEx returned a *Ticket, but CreateKey returns Ticket)

attest/tpm_linux.go
DenisKarch: Probably yes, should we also drop the check for KeyEncoding when loading?

attest/tpm_linux.go
twitchy-jsonp: Got it, cheers.

attest/tpm_linux.go
twitchy-jsonp: The intent of the `KeyEncoding` field is to describe how the serialized key is represented. `KeyEncodingParameterized` should indicate to `LoadKey` that the key is a primary key that needs to be generated to get the key handle. `KeyEncodingEncrypted` symbolizes that the encrypted private blob is present, and the key needs to be loaded to get the handle.

The aim of this was to support keys that were generated with an older version of the library.

So either:

1. We should change the check of KeyEncoding to `KeyEncodingEncrypted`, dropping support for loading the old key format
2. We should use a switch to support loading both.

My preference is the second, but its up to you.

DenisKarch:
don't pull this in yet btw. waiting for go-tpm to approve a change for CreateKey https://github.com/google/go-tpm/pull/85, will update the go.mod when that is sorted.

attest/tpm_linux.go
DenisKarch: I've already made the change for the 1st option

twitchy-jsonp:
Sounds good, lets merge this when the dependencies are ready.

PR #13, opened 2019-04-18
Import verifier library into go-attestation.

twitchy-jsonp:
PTAL. Also, can you double-check what I am doing with the proto files is sane?


proto/tpm.proto
```diff
@@ -0,0 +1,113 @@
+syntax = "proto3";
+
+package goattestation.proto;
```
brandonweeks: Nit: "package go_attestation;"

PR #16, opened 2019-04-24
Extract and return PCR digest for Quote on TPM1.2

DenisKarch:
go-tspi and go-tpm return different values for "quote".
In both cases we want the PCR digest so we extract it from the data returned.

### DIFF
--- a/attest/tpm_linux.go
+++ b/attest/tpm_linux.go
@@ -341,14 +341,21 @@ func (k *Key) ActivateCredential(t *TPM, in EncryptedCredential) ([]byte, error)
 }
 
 func (k *Key) quote12(ctx *tspi.Context, nonce []byte) (*Quote, error) {
-	quote, rawSig, err := attestation.GetQuote(ctx, k.KeyBlob, nonce)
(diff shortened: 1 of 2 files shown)


attest/tpm_linux.go
```diff
@@ -341,10 +341,12 @@ func (k *Key) ActivateCredential(t *TPM, in EncryptedCredential) ([]byte, error)
 }
 
 func (k *Key) quote12(ctx *tspi.Context, nonce []byte) (*Quote, error) {
-	quote, rawSig, err := attestation.GetQuote(ctx, k.KeyBlob, nonce)
+	quoteInfo, rawSig, err := attestation.GetQuote(ctx, k.KeyBlob, nonce)
 	if err != nil {
 		return nil, fmt.Errorf("GetQuote() failed: %v", err)
 	}
+	// go-tspi returns TPM_QUOTE_INFO. We only want the digest of the PCRs
+	quote := quoteInfo[8:28]
```
twitchy-jsonp: Could we use tpmutil.Unpack for this?

```go
var version uint32
var tag [4]byte
var digest [20]byte
if err := tpmutil.Unpack(in, &version, &tag, &digest); err != nil {
  ...
}
```

PR #53, opened 2019-07-19
Refactor part 1: Refactor logic for keys into structs for each TPM/platform invariant.

twitchy-jsonp:


attest/attest.go
```diff
@@ -52,12 +53,12 @@ type OpenConfig struct {
...
 
-// KeyEncoding indicates how an exported TPM key is represented.
-type KeyEncoding uint8
+// keyEncoding indicates how an exported TPM key is represented.
+type keyEncoding uint8
 
 // Key encodings
 const (
-	KeyEncodingInvalid KeyEncoding = iota
+	KeyEncodingInvalid keyEncoding = iota
```
ericchiang: nit: the 'KeyEncoding...' variables should also be private.

attest/attest.go
```diff
@@ -77,6 +78,49 @@ const (
...
+}
+
+// ActivateCredential decrypts the specified credential using the key.
+// This operation is synonymous with TPM2_ActivateCredential.
+func (k *AIK) ActivateCredential(tpm *TPM, in EncryptedCredential) ([]byte, error) {
+	return k.aik.ActivateCredential(tpm, in)
+}
+
+// Quote returns a quote over the platform state, signed by the AIK.
+func (k *AIK) Quote(tpm *TPM, nonce []byte, alg tpm2.Algorithm) (*Quote, error) {
```
ericchiang: Why is alg `tpm2.Algorithm`? Can't this be a tpm1.2 TPM?

Maybe have this package define its own Algorithm type?

attest/attest.go
```diff
@@ -111,6 +155,17 @@ type PlatformEK struct {
 	Public crypto.PublicKey
 }
 
+// AIKParameters describes information about an AIK. This information
+// is typically used to generate an activation challenge.
+type AIKParameters struct {
+	Version TPMVersion
+
+	Public            []byte
```
ericchiang: Can you add a link and comments explaining what these fields are?

attest/attest.go
```diff
@@ -111,6 +155,17 @@ type PlatformEK struct {
 	Public crypto.PublicKey
 }
 
+// AIKParameters describes information about an AIK. This information
+// is typically used to generate an activation challenge.
+type AIKParameters struct {
+	Version TPMVersion
```
ericchiang: I don't like that there are multiple ways to get the TPMVersion, can you drop this and continue to have users call `func (t *TPM) Info()`?

attest/storage.go
```diff
@@ -0,0 +1,61 @@
...
+// the License.
+
+package attest
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// serializedKey represents a loadable, TPM-backed key.
```
ericchiang: nit: it might be helpful to indicate which of these fields are relevant to each key type (1.2 vs 2.0, linux vs windows)

attest/tpm_linux.go
```diff
@@ -422,62 +309,64 @@ func (t *TPM) MintAIK(opts *MintOptions) (*Key, error) {
...
+			},
 		}, nil
 
 	default:
 		return nil, fmt.Errorf("unsupported TPM version: %x", t.version)
 	}
 }
 
-// LoadKey loads a previously-created key into the TPM for use.
+// LoadAIK loads a previously-created aik into the TPM for use.
```
ericchiang: nit: mention that this takes the result of AIK.Serialize()

attest/tpm_linux.go
```diff
@@ -422,62 +309,64 @@ func (t *TPM) MintAIK(opts *MintOptions) (*Key, error) {
...
-		return nil, errors.New("key TPM version does not match opened TPM")
+func (t *TPM) LoadAIK(opaqueBlob []byte) (*AIK, error) {
+	sKey, err := deserializeKey(opaqueBlob, t.version)
+	if err != nil {
+		return nil, fmt.Errorf("deserializeKey() failed: %v", err)
 	}
-	if k.Purpose != AttestationKey {
-		return nil, fmt.Errorf("unsupported key kind: %x", k.Purpose)
+	if sKey.Encoding != KeyEncodingEncrypted {
+		return nil, fmt.Errorf("unsupported key encoding: %x", sKey.Encoding)
```
ericchiang: nit: keyEncoding should implement Stringer (maybe a TODO?)

attest/tpm_windows.go
```diff
@@ -410,42 +304,73 @@ func (t *TPM) MintAIK(opts *MintOptions) (*Key, error) {
...
+	default:
+		return nil, fmt.Errorf("cannot handle TPM version: %v", t.version)
 	}
-	if k.KeyEncoding != KeyEncodingOSManaged {
-		return nil, fmt.Errorf("unsupported key encoding: %x", k.KeyEncoding)
+}
+
+// LoadAIK loads a previously-created AIK into the TPM for use.
+// A key loaded via this function needs to be closed with .Close().
+func (t *TPM) LoadAIK(opaqueBlob []byte) (*AIK, error) {
```
ericchiang: nit: can you make this exported function defined in a file without any build tags? Then immediately call out to a private method that each arch implements?

```
func (t *TPM) LoadAIK(opaqueBlob []byte) (*AIK, error) {
   return t.loadAIK(opaqueBlob)
}
```

That way you can write the godoc once and it won't get out of sync

attest/attest.go
twitchy-jsonp: Info() results in a roundtrip to the TPM to retrieve a whole bunch of fields, so I would rather not do that for trivial fields like version.
Maybe we can have a `tpm.Version()` instead, given that TPMVersion is a field on the TPM struct?

PR #57, opened 2019-07-23
Update TCG vendorID list

twitchy-jsonp:


attest/vendors.go
```diff
@@ -38,6 +40,7 @@ var vendors = map[TCGVendorID]string{
 	1415073280: "Texas Instruments",
 	1464156928: "Winbond",
 	1380926275: "Fuzhou Rockchip",
+	1196379975: "Google",
```
ericchiang: This doesn't align with any of the values in the linked document. All I see is `0x6666` and `0x47 0x4F 0x4F 0x47`

twitchy-jsonp: This is the same value, just decimal encoded.

```shell

[jsonp@jsonp]:~> python -c "print([hex(0xff & (1196379975 >> (a*8))) for a in xrange(4)])"
['0x47', '0x4f', '0x4f', '0x47']
```

ericchiang: Thanks!

PR #58, opened 2019-07-23
Make SRK/AIK templates consistent with go-tpm-tools

twitchy-jsonp:


twitchy-jsonp:
Friendly pingaling :)

josephlr:
This isn't quite right for the SRK, we need to set `ModulusRaw: make([]byte, 256)` just like we do for the EK. It also might be a good idea to update the `go-tpm` dependency to the latest master, to get https://github.com/google/go-tpm/pull/120.

For the AIK, we don't have to have an exact match, as these are not persisted at a standard handle. However, the AIK in `go-attestation` differs from the one in `go-tpm-tools` in the following ways:
  - The template Modulus is different
    - `go-attestation` uses `Modulus: big.NewInt(0)` (i.e. an zero-length `ModulusRaw`)
    - `go-tpm-tools` uses a 256 byte nonce for the initial `ModulusRaw`, so that multiple AIKs can be easily issued.
  - The parent is different
    - `go-attestation` uses `Create` to make the AIK a child key of the SRK. This means two RSA keys have to be loaded to mint the AIK.
    - `go-tpm-tools` uses `CreatePrimary` to make the AIK as a primary object in the owner hierarchy

Alternatively, you could just depend on go-tpm-tools to get the default templates (but that might not be desirable right now, and that's OK).

twitchy-jsonp:
WRT `CreatePrimary`, we used to do that as well but found we had the multi-second performance hit to generate a key each time. Instead, creating a leaf key in the owner hierarchy allows us to load & unload the key, and better matches what Windows is doing.

Will the empty `RSAParams` value encode exponent as `defaultRSAExponent` ?

twitchy-jsonp:
Is there a spec or convention around the parameters of the AIK or SRK? trying to work out the implications of having `unique` set or not.

josephlr:
For the SRK, there is no standard but there is a recommendation based on the corresponding EK template. You take the EK template and make the following changes:
  - AuthPolicy is empty
  - NoDA is set
  - AdminWithPolicy is unset
  - UserWithAuth is set

This is (a) what windows does, and (b) what was perprovisioned on all the real TPM2s I could find.

We don’t want an empty RSA Parma’s, as you need to encode algorithms and key length.

As for the AIK, the specifics matter a lot less, as they can all exist in parallel just fine. We only set unique so that we can generate multiple different AIKs. This is also how the AIKs on GCE VMs work (with a random nonce).

josephlr:
Also if you’re concerned about time, you could just persist the AIK handle as well (this seems common in TPMs) to something like

attest/tpm.go
```diff
@@ -46,28 +45,26 @@ var (
...
-		Attributes: tpm2.FlagSignerDefault | tpm2.FlagNoDA,
+		Attributes: tpm2.FlagSignerDefault,
 		RSAParameters: &tpm2.RSAParams{
 			Sign: &tpm2.SigScheme{
 				Alg:  tpm2.AlgRSASSA,
 				Hash: tpm2.AlgSHA256,
 			},
-			KeyBits: 2048,
-			Modulus: big.NewInt(0),
+			KeyBits:    2048,
```
josephlr: Having a 256-byte random `ModulusRaw` might be a good idea here, it will prevent two applications from inadvertently creating the same AIK. If you don't care about that, leaving `ModulusRaw` empty is the right choice (and is what `tpm2-tools` does.

twitchy-jsonp: SGTM about leaving it empty.

twitchy-jsonp:
> I think both are wrong, they should be child keys of the EK.

I think this is a matter of perspective. Following the TCG vision where you can have an endorser which is a separate entity to some service provider, it makes sense to have the AIK in a separate (endorsement) hierarchy so it survives across the lifecycle of any storage key, and remains conceptually separate to 'owner' use of keys.

Putting it in the endorsement hierarchy is useful as well in that only keys chaining to EK get un-obfuscated values for certain fields (i think `clock` is one of them) for privacy reasons.

However, having the AIK in the storage hierarchy is useful when you want to bind the lifecycle of your AIK to that of your SRK, which usually corresponds to a deployment of a managed image or an identity. This is how windows does it (coincidentally corresponding strongly with bitlocker) and a requirement (chain to SRK) we received from a platform team internally.

attest/tpm.go
```diff
@@ -46,28 +45,26 @@ var (
...
+		Attributes: tpm2.FlagStorageDefault | tpm2.FlagNoDA,
 		RSAParameters: &tpm2.RSAParams{
 			Symmetric: &tpm2.SymScheme{
 				Alg:     tpm2.AlgAES,
 				KeyBits: 128,
 				Mode:    tpm2.AlgCFB,
 			},
-			KeyBits: 2048,
-			Modulus: big.NewInt(0),
+			KeyBits:    2048,
```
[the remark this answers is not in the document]
twitchy-jsonp: Done.

PR #60, opened 2019-07-30
attest: remove unused KeyPurpose exported type

ericchiang:
Was going through the godoc and it's not clear that this is ever used.
To clean up the API, remove KeyPurpose for now. This could probably be
an internal validation anyway, right?

@twitchy-jsonp please wait to get back from OOO before reviewing. Not going to merge anything without your review.

### DIFF
--- a/attest/attest.go
+++ b/attest/attest.go
@@ -83,17 +83,6 @@ const (
 	keyEncodingParameterized
 )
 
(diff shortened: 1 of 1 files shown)


twitchy-jsonp:
Yeah this is not ever used, but will make sense if we add support for other TPM keys / operations.

ericchiang:
Cool, we can add it back then :)

PR #62, opened 2019-07-30
attest: move public key parsing server side

ericchiang:
Event log parsing requires knowning both the public key and signing
parameters. Symmantically, this information should be from an
attested public key blob, not additional data passed by the client.

Move the public key parsing to the AttestationParameters instead of the
AIK. Alternative APIs might be:

```go
func (a *AttestationParameters) PublicKey() (crypto.PublicKey, error)
func (a *AttestationParameters) SignerOpts() (crypto.SignerOpts, error)
```

Or:

```go
type PublicKey struct {
    Key  crpyto.PublicKey
    Opts crypto.SignerOpts
}
func (a *AttestationParameters) PublicKey() (*PublicKey, error)
```

Or:

```
type PublicKey struct {
    Key  crpyto.PublicKey
    Opts crypto.SignerOpts
}
func ParsePublicKey(blob []byte) (*PublicKey, error)
```

Open for discussion

### DIFF
--- a/attest/attest_simulated_tpm20_test.go
+++ b/attest/attest_simulated_tpm20_test.go
@@ -153,6 +153,21 @@ func TestSimTPM20ActivateCredential(t *testing.T) {
 	}
 }
 
+func TestParseAIKPublic20(t *testing.T) {
+	sim, tpm := setupSimulatedTPM(t)
+	defer sim.Close()
+
+	aik, err := tpm.MintAIK(nil)
+	if err != nil {
+		t.Fatalf("MintAIK() failed: %v", err)
+	}
+	defer aik.Close(tpm)
+	params := aik.AttestationParameters()
+	if _, err := ParseAIKPublic(TPMVersion20, params.Public); err != nil {
+		t.Errorf("parsing AIK public blob: %v", err)
+	}
+}
+
 func TestSimTPM20Quote(t *testing.T) {
 	sim, tpm := setupSimulatedTPM(t)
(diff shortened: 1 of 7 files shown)


attest/attest_simulated_tpm20_test.go
```diff
@@ -153,6 +153,21 @@ func TestSimTPM20ActivateCredential(t *testing.T) {
 	}
 }
 
+func TestParseAIKPublic(t *testing.T) {
```
twitchy-jsonp: Could you add a test for TPM 1.2 as well? maybe just parse the public handled in `TestMintAIK` (attest_tpm12_test.go) or something?

ericchiang: Added a test. ptal

ericchiang: Also, I don't have a good idea how to run these tests. My guess is that it's not part of CI?

twitchy-jsonp: The CI runs them: https://source.cloud.google.com/results/invocations/c3647263-0de6-4b0d-9a7b-96e3c0c3490a/targets/attest/tests;group=attest;test=TestParseAIKPublic12;row=5

Getting them to run on your machine is witchcraft though: https://github.com/google/go-attestation/blob/master/ci/setup_tpm12_simulator.sh

ericchiang: Sweet :)

PR #64, opened 2019-08-01
attest: add event log parsing logic

ericchiang:
This PR adds event log parsing logic. It's main goal is to require
validation at the same time as parsing, so structured events are always
verified against a quote. This new API replaces the existing "verifier"
package.

It's not a goal of this PR to parse the event data. This will be a
follow up, but since different users might want to parse different
events based on the OS, this API lets users of this package implement
custom event data parsing if they absolutely need to.

TODO:
- [x] Implement TPM 1.2 logic
- [x] Add tests once #62 is merged

Updates #9


attest/eventlog.go
```diff
@@ -0,0 +1,399 @@
...
+
+	"github.com/google/go-tpm/tpm2"
+	"github.com/google/go-tpm/tpmutil"
+)
+
+// TPM algorithms. See the TPM 2.0 specification section 6.3.
+//
+// https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-2-Structures-01.38.pdf#page=42
+const (
+	algSHA1   uint16 = 0x0004
```
twitchy-jsonp: Should/could we use the constants from tpm2?

ie: https://github.com/google/go-attestation/blob/a801f7333b9e043c39ea938c44884b05fab9691f/verifier/quote.go#L18

attest/eventlog.go
```diff
@@ -0,0 +1,399 @@
...
+		return nil, fmt.Errorf("parsing measurement log: %v", err)
+	}
+	events, err = replayEvents(rawEvents, pcrs)
+	if err != nil {
+		return nil, fmt.Errorf("pcrs failed to replay")
+	}
+	return events, nil
+}
+
+type rawAttestationData struct {
```
twitchy-jsonp: This seems TPM 1.2 specific?

ericchiang: Yep, renaming to `rawQuoteInfo12` and adding links to the spec :)

attest/eventlog.go
ericchiang: Changing this to use tpm2.Algorithm. Going to keep the switch statement instead of using tpm2.Algorithm.HashConstructor(). If someone is using something other than SHA1/SHA256 I want to know and purposefully add it.

ericchiang:
@twitchy-jsonp to add a hash argument to PCR for #72 

@brandonweeks to add a method for determining the supported hash of a measurement log.

Maybe modify MeasurementLog to:

```
func (t *tpm) MeasurementLog() ([]byte, HashAlg, error)
```

PR #67, opened 2019-08-07
Add firmware version to TPMInfo for TPM 2.0 devices.

twitchy-jsonp:
Not 100% on the API for this, PTAL.

### DIFF
--- a/attest/attest.go
+++ b/attest/attest.go
@@ -224,16 +224,17 @@ var (
 
 // TPMInfo contains information about the version & interface
 // of an open TPM.
-// FWVersionMajor and FWVersionMinor describe the firmware version of
-// the TPM, but are only available for TPM 2.0 devices.
 type TPMInfo struct {
 	Version      TPMVersion
 	Interface    TPMInterface
 	VendorInfo   string
 	Manufacturer TCGVendorID
 
-	FWVersionMajor int
-	FWVersionMinor int
+	// FirmwareVersionMajor and FirmwareVersionMinor describe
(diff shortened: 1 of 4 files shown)


attest/attest.go
```diff
@@ -224,11 +224,16 @@ var (
 
 // TPMInfo contains information about the version & interface
 // of an open TPM.
+// FWVersionMajor and FWVersionMinor describe the firmware version of
```
ericchiang: Comment should probably be over the fields not the struct?

Are there any links or background you can add to this comment? E.g. what are reasonable values to take? Could these potentially indicate vulnerable TPMs?

attest/attest.go
```diff
@@ -224,11 +224,16 @@ var (
...
+// the TPM, but are only available for TPM 2.0 devices.
 type TPMInfo struct {
 	Version      TPMVersion
 	Interface    TPMInterface
 	VendorInfo   string
 	Manufacturer TCGVendorID
+
+	FWVersionMajor int
```
ericchiang: nit: Maybe don't abbreviate? These are going to be used sufficiency rare and I don't get FW -> Firmware

```
FirmwareVersionMajor int
FirmwareVersionMinor int
```

attest/attest.go
twitchy-jsonp: Done. Not really, the values are vendor specific.

attest/attest.go
twitchy-jsonp: Done.

PR #69, opened 2019-08-12
WIP - example client

twitchy-jsonp:
I'm working on making an example but want to get your thoughts on the layout.

Main question: Do you think its worthwhile to build an end-to-end example, including network RPCs? Or maybe its acceptable to do the approach below where I dump the values to stdout and then you can copy paste them into the 'server' (maybe this can be a webui rather than a command line binary) for verification.

### DIFF
--- a/attest/example_test.go
+++ b/attest/example_test.go
@@ -129,3 +129,48 @@ func TestExampleAIK(t *testing.T) {
 	ExampleAIK_credentialActivation()
 	ExampleAIK_quote()
 }
+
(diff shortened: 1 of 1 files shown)


twitchy-jsonp:
I've refactored & implemented credential activation + attestation, in a HTTP-based client-server demo. PTAL when you are feeling better :)

twitchy-jsonp:
It's likely we'll persue the phoneapp-laptop route or simply use attest-tool.

Issue #75, opened 2019-08-20
attest: figure out hashing algorithm strategy


ericchiang:
This is a super-set of https://github.com/google/go-attestation/issues/72

It's currently unclear when a user needs to provide SHA1 or SHA256, or when we can infer it from other data. There are several places where we need a hash:

* To determine the signature algorithm on the quote. Is this always inferred from the TPM public blob?
* To determine which quote we should send to the server. TPM 2.0 doesn't guarantee that the eventlog will correctly contain SHA256, even if the PCRs store SHA256 values. Should this be determined by inspecting the measurement log?
* It's unclear why AIK.Quote() requires a HashAlg, or how this should be filled.

For now, I propose we modify the MeasurementLog signature to return the hash:

```
func (t *TPM) MeasurementLog() ([]byte, HashAlg, error)
```

And the PCRs call to take a hash:

```
func (t *TPM) PCRs(alg HashAlg) ([]PCR, error)
```

Users would then do something like

```
rawLog, alg, err := tpm.MeasurementLog()
if err != nil {
    // ...
}
pcrs, err := tpm.PCRs(alg)
```

twitchy-jsonp:
According to https://www.trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-3-Commands-01.38.pdf, section 18.4:

The provided algorithm is used to digest the PCRs into a single value for the signature. In theory, TPM 2 devices will always support SHA256, but iirc I had some issues getting SHA256 quotes working for non-sha256 event logs (but my memory is flaky).

ericchiang:
brandonweeks@ has been volunteered to take this

PR #76, opened 2019-08-20
#75: Make PCRs() take the digest algorithm

### LINKED ISSUE #72: PCRs method doesn't allow providing hash algorithm

twitchy-jsonp:
Closes #72
Updates #75

### DIFF
--- a/attest/tpm_linux.go
+++ b/attest/tpm_linux.go
@@ -17,7 +17,6 @@
 package attest
 
 import (
-	"crypto"
 	"crypto/rsa"
 	"encoding/binary"
 	"errors"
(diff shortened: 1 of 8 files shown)


attest/tpm_linux.go
```diff
@@ -359,48 +358,40 @@ func allPCRs12(ctx *tspi.Context) (map[uint32][]byte, error) {
 }
 
 // PCRs returns the present value of all Platform Configuration Registers.
-func (t *TPM) PCRs() (map[int]PCR, tpm2.Algorithm, error) {
+func (t *TPM) PCRs(alg HashAlg) ([]PCR, error) {
```
ericchiang: nit: godoc needs to be updated.

It's also kinda weird that we're defining public APIs in _linux.go and _windows.go files. Maybe add a TODO for something similar to key_{linux,windows}.go where the public API lives in a file without build tags, and the OS specific files only contain private methods?

ericchiang:
My comment was resolved, but it doesn't look like the function comment for `func (t *TPM) PCRs(alg HashAlg) ([]PCR, error)` was updated to explain what the alg is? Was that a mistake? :)

twitchy-jsonp:
I'm not sure if github is wrong or if my git is messed up, look after the talk?

PR #77, opened 2019-08-20
#75: Make PCRs() take the digest algorithm

twitchy-jsonp:


twitchy-jsonp:
This is #76 except a new PR as the old one did not appear to be updating the preview for some reason.

Issue #85, opened 2019-08-27
attest: Allow verification of quote without a corresponding event log


twitchy-jsonp:
Users may choose to use Quote() to challenge the presence of an AIK, by quoting over PCRs and providing a nonce. In this case, there is no event log (and there may be none present), but we still want to be able to verify the quote.

Currently, the API for verifying a quote is coupled into event log verification (users must call `EventLog{}.Validate()`, which will fail if they do not provide an event log).

Proposal 1: Add a `Verify(aik, nonce, pcrs)` method directly on `Quote`
Proposal 2: Make it legal to call `Validate` on `EventLog` where no event log is present.

WDYT @ericchiang ?

ericchiang:
Closed by #87

PR #87, opened 2019-08-28
#85 Separate quote verification from EventLog API

twitchy-jsonp:
Implements API discussed in #85 

 - Quote/AIK verification removed from `EventLog.Validate()`, moved into new method `AIKPublic.Verify()`
 - Moved AIKPublic & associated verification logic into own file & test file
 - Separate AIK verification logic from event log test & put in its own test
 - Added missing license headers.


ericchiang:
lgtm. please squash your commits before merging.

PR #93, opened 2019-09-04
Document overview of the credential activation procedure

twitchy-jsonp:


twitchy-jsonp:
I've tried to shop this to be more concrete in the description of the process, and use the EK-AK scenario at the beginning.

I made up anchor key because I couldn't find any TCG terminology for that key (the command just calls it keyHandle).

PR #94, opened 2019-09-04
internal/eventlog: add code for parsing secure boot variables

ericchiang:
This is being prototyped in an internal package as we start to open
source. This code will either live in attest, or in a separate eventlog
package in the future.


ericchiang:
Updates #9

PR #96, opened 2019-09-06
docs: Add appendix about how activation challenge is generated

twitchy-jsonp:


ericchiang:
Hey Tom, can you describe the intended audience for this document? This is good information, but some of this feel more appropriate as inline code comments or references to the spec. What's the technical level you're aiming at? Someone trying to use the package? Someone trying to implement attestation themselves?

Also, the packet diagram doesn't render for me :)

image

twitchy-jsonp:
The is intended as a discoverable resource to anyone trying to implement or debug credential activation. I wrote this as existing resources and the spec are either vague, wrong, or hard to understand on this process (it took over 2 weeks to get this working when writing the implementation, which would not be the case if I wasnt having to do a lot of guessing + cross-referencing various broken implementations).

Not sure where the right place for this is. It feels like a blog post would be ideal but I'm not sure exactly where that would go. I don't think inline comments would be appropriately discoverable thou.

PR #97, opened 2019-09-09
Continuation of #96

twitchy-jsonp:
This is a continuation of the discussion from #96 - doesnt look like I can re-open a PR once the commit it last mentions got merged.

Sorry for messing up the branch. This PR is based off the orphaned commit.


twitchy-jsonp:
Bringing in 1st comment from previous PR from Eric:

> Hey Tom, can you describe the intended audience for this document? This is good information, but some of this feel more appropriate as inline code comments or references to the spec. What's the technical level you're aiming at? Someone trying to use the package? Someone trying to implement attestation themselves?

> Also, the packet diagram doesn't render for me :)

twitchy-jsonp:
Bringing in my reply:

> The is intended as a discoverable resource to anyone trying to implement or debug credential activation. I wrote this as existing resources and the spec are either vague, wrong, or hard to understand on this process (it took over 2 weeks to get this working when writing the implementation, which would not be the case if I wasnt having to do a lot of guessing + cross-referencing various broken implementations).

> Not sure where the right place for this is. It feels like a blog post would be ideal but I'm not sure exactly where that would go. I don't think inline comments would be appropriately discoverable thou.

PR #101, opened 2019-09-19
attest: don't define exported API in files with build tags

ericchiang:
It's best practice to define as much code, especially exported API, in
files that can build on any platform. With as little code as possible in
OS specific files.

Ensure files with build tags don't contain any exported APIs. This helps
us not accidentally define API that only works on one platform, or have
incompatible method defintions between OSes.

There are no functional changes to attest's exported methods.

TODO: follow up with an "unsupported" implementation so this builds on
Mac or without CGO (e.g. for servers)?


attest/key_windows.go
```diff
@@ -168,19 +156,19 @@ func (k *key20) Quote(t *TPM, nonce []byte, alg HashAlg) (*Quote, error) {
...
+func (k *key20) close(tpm *platformTPM) error {
 	return closeNCryptObject(k.hnd)
 }
 
 // Delete permenantly removes the key from the system. This method
 // invalidates Key and any further method invocations are invalid.
-func (k *key20) Delete(tpm *TPM) error {
+//
+// TODO(ericchiang): this isn't exported. remove the method or do we need it?
+func (k *key20) Delete(tpm *platformTPM) error {
```
ericchiang: @twitchy-jsonp fyi

twitchy-jsonp: I added it in for a one-time cleanup, and it only makes sense on Windows. We can remove it.

ericchiang:
This all builds

```
$ GOOS=linux go build github.com/google/go-attestation/attest
$ GOOS=linux go test -c github.com/google/go-attestation/attest
$ GOOS=windows go build github.com/google/go-attestation/attest
$ GOOS=windows go test -c github.com/google/go-attestation/attest
```

Issue #63, opened 2019-07-31
Naming cleanups


ericchiang:
Looking at the exported API the package is a little cluttered. https://godoc.org/github.com/google/go-attestation/attest

Some thoughts for cleaning up the API, these are mostly just nits:
* Replace HashAlg with crypto.Hash
* PlatformEK -> EK
* Remove KeyPurpose (#60)
* MintOptions -> AIKConfig / TPM.MintAIK -> TPM.NewAIK (so all the AIK options are ordered together)

ericchiang:
Other thoughts:
* Can TPMInterface be private? We only need to determine how to load the AIK.
* Convert `map[int]PCR` to `[]PCR`? The PCR struct already contains an index, and it's redundant to hold that information as a map key as well.

ericchiang:
cc @twitchy-jsonp

PR #108, opened 2019-09-20
Add an initial implementation of parsing event types

mjg59:
Raw events aren't enormously helpful. Add an initial implementation of
some of the basic event types - this is largely a scratch proposal so we
can figure out API design.


ericchiang:
Can you please put this in a different package for a bit? This would double the amount of exported API and we've worked hard to get the godoc to be digestible (https://github.com/google/go-attestation/issues/63): https://godoc.org/github.com/google/go-attestation/attest

With that in mind, I think we can avoid making the user type cast every event back to the type we want. Tom and I were discussing having Parse methods that parsed the event body. e.g.

```
type MicrocodeEvent struct {
    // ...
}
func ParseMicrocodeEvent(eventData []byte) (*MicrocodeEvent, error) {
    // ...
}
```

So the user can iterate through the events and use the corresponding Parse method

```
for _, e := range events {
    switch e.Type {
    case eventlog.TypeCPUMicrocode:
        mcEvent, err := eventlog.ParseMicrocodeEvent(e.Data)
    // ... 
    }
}
```

mjg59:
> Can you please put this in a different package for a bit? This would double the amount of exported API and we've worked hard to get the godoc to be digestible (#63): https://godoc.org/github.com/google/go-attestation/attest

Sure, though that'll mean exporting rawEvent?

> With that in mind, I think we can avoid making the user type cast every event back to the type we want. Tom and I were discussing having Parse methods that parsed the event body. e.g.

> for _, e := range events {
>     switch e.Type {
>     case eventlog.TypeCPUMicrocode:
>         mcEvent, err := eventlog.ParseMicrocodeEvent(e.Data)

Is this really easier than
```
switch e.Type {
  case eventlog.TypeCPUMicrocode:
     mcEvent, ok := e.(CPUMicrocodeEvent)
```
? Splitting out the individual parse methods means a lot of basically identical functions, which feels a little less clean than basically doing the same in a switch statement (although for more complex types, we should probably also be exporting information about whether we succeeded in parsing the event)

ericchiang:
You can use the exported Event type :) https://godoc.org/github.com/google/go-attestation/attest#Event

I think both strategies are similar amounts of code for the caller. For example to get a single event:

```go
events, err := eventlog.ParseEvents(verifiedEvents)
if err != nil {
    // ...
}
for _, e := range events {
    if pbce, ok := e.(eventlog.PrebootCertEvent); ok {
        // use parsed event
    }
}
```

vs.

```go
for _, e := range events {
    if e.Type == eventlog.TypePrebootCertEvent {
        pbce, err := eventlog.ParsePrebootCertEvent(e.Data)
        if err != nil {
            // ...
        }
        // use parsed event
    }
}
```

I suggested the latter because there's a clearer place for users to parse event types that we don't provide a parser for.

Either way, I don't think there's a huge difference. Feel free to go with either API

mjg59:
> You can use the exported Event type :) https://godoc.org/github.com/google/go-attestation/attest#Event

But EventLog contains an array of rawEvent rather than Event?

ericchiang:
Part of the API design is to require replaying the event log against PCRs before moving forward https://godoc.org/github.com/google/go-attestation/attest#EventLog.Verify

We want users of the API to be safe by default, and not mistakenly use something that's not been verified.

mjg59:
Oh, yes, I see what you mean now.

mjg59:
Ok, added support for Windows event types and tests against a bunch of sample event logs.

events/events.go
```diff
@@ -0,0 +1,789 @@
...
+}
+
+func parseEfiVariableData(b []byte, parsedEvent *uefiVariableEvent) error {
+	var header uefiVariableDataHeader
+	buf := bytes.NewBuffer(b)
+	err := binary.Read(buf, binary.LittleEndian, &header)
+	if err != nil {
+		return err
+	}
+	unicodeName := make([]uint16, header.UnicodeNameLength)
```
ericchiang: same dos

events/devicepath.go
```diff
@@ -0,0 +1,586 @@
...
+func dumpEfiDevicePath(buf io.Reader, dp efiDevicePathHeader, prefix string) string {
+	data := make([]byte, dp.Length-4)
+	binary.Read(buf, binary.LittleEndian, &data)
+	if prefix == "" {
+		return fmt.Sprintf("Path(%d,%d,%02x)", dp.Type, dp.SubType, data[:])
+	}
+	return fmt.Sprintf("%s(%d,%02x)", prefix, dp.SubType, data[:])
+}
+
+// efiDevicePath translates an EFI Device Path into the canonical string representation
```
ericchiang: got a link for where the canonical representation is specified?

liamjm:
I'm going to close this PR as obsolete, please re-open if you want to continue this.

PR #110, opened 2019-09-24
Expose information about a failing replay via ReplayErr structure

twitchy-jsonp:


attest/eventlog.go
```diff
@@ -77,6 +88,9 @@ type EventLog struct {
 func (e *EventLog) Verify(pcrs []PCR) ([]Event, error) {
 	events, err := replayEvents(e.rawEvents, pcrs)
 	if err != nil {
+		if _, isReplayErr := err.(ReplayError); isReplayErr {
+			return nil, err
+		}
 		return nil, fmt.Errorf("pcrs failed to replay: %v", err)
```
ericchiang: Use error wrapping? https://golang.org/pkg/errors/#pkg-overview

attest/eventlog.go
```diff
@@ -31,6 +31,17 @@ import (
...
 )
 
+// ReplayError describes the parsed events that failed to verify against
+// a particular PCR.
+type ReplayError struct {
+	BadEvents   []Event
+	InvalidPCRs []int
+}
+
+func (e ReplayError) Error() string {
```
ericchiang: nit: for lint please add a godoc comment here

attest/eventlog.go
```diff
@@ -31,6 +31,17 @@ import (
 	"github.com/google/go-tpm/tpmutil"
 )
 
+// ReplayError describes the parsed events that failed to verify against
+// a particular PCR.
+type ReplayError struct {
+	BadEvents   []Event
+	InvalidPCRs []int
```
ericchiang: Events already contain the PCR index, no need to duplicate?

attest/eventlog.go
```diff
@@ -31,6 +31,17 @@ import (
 	"github.com/google/go-tpm/tpmutil"
 )
 
+// ReplayError describes the parsed events that failed to verify against
+// a particular PCR.
+type ReplayError struct {
+	BadEvents   []Event
```
ericchiang: nit: this could probably just be "Events" right? "Bad" is already implied since it's an error.

attest/eventlog.go
twitchy-jsonp: Do we want to require all our users to have Go 1.13? Seems a little early?

attest/eventlog.go
twitchy-jsonp: Done.

attest/eventlog.go
twitchy-jsonp: Done.

attest/eventlog.go
twitchy-jsonp: Done. I've kept invalidPCRs around as unexported to keep the generation of the error string simple.

attest/eventlog.go
ericchiang: Yeah you're probably right.

PR #112, opened 2019-09-25
Re-work eventlog replay algorithm

twitchy-jsonp:
- Support replaying against a set of PCRs which includes multiple values for the same index
 - Consider a specific PCR 'verified' if at least one provided PCR value matched all events
 - Verify `tpm20GeneratedMagic` in `validate20Quote()` to prevent use of `TPM2_Sign` to fake quotes

### DIFF
--- a/attest/eventlog.go
+++ b/attest/eventlog.go
@@ -57,6 +57,9 @@ type EventType uint32
 // Event is a single event from a TCG event log. This reports descrete items such
 // as BIOs measurements or EFI states.
 type Event struct {
+	// order of the event in the event log.
+	sequence int
+
 	// PCR index of the event.
 	Index int
 	// Type of the event.
@@ -83,9 +86,10 @@ type EventLog struct {
 	rawEvents []rawEvent
 }
 
-// Verify replays the event log against a TPM's PCR values, returning events
-// from the event log, or an error if the replayed PCR values did not match the
-// provided PCR values.
+// Verify replays the event log against a TPM's PCR values, returning the
+// events which could be matched to a provided PCR value.
+// An error is returned if the replayed digest for events with a given PCR
+// index do not match any provided value for that PCR index.
 func (e *EventLog) Verify(pcrs []PCR) ([]Event, error) {
 	events, err := replayEvents(e.rawEvents, pcrs)
 	if err != nil {
@@ -224,10 +228,10 @@ var hashBySize = map[int]crypto.Hash{
 	crypto.SHA256.Size(): crypto.SHA256,
 }
 
-func extend(pcr, replay []byte, e rawEvent) ([]byte, Event, error) {
+func extend(pcr, replay []byte, e rawEvent) (pcrDigest []byte, eventDigest []byte, err error) {
 	h, ok := hashBySize[len(pcr)]
 	if !ok {
-		return nil, Event{}, fmt.Errorf("pcr %d was not a known hash size: %d", e.index, len(pcr))
+		return nil, nil, fmt.Errorf("pcr %d was not a known hash size: %d", e.index, len(pcr))
 	}
 	for _, digest := range e.digests {
 		if len(digest) != len(pcr) {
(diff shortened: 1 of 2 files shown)


attest/eventlog.go
```diff
@@ -241,46 +249,88 @@ func extend(pcr, replay []byte, e rawEvent) ([]byte, Event, error) {
...
+			return nil, fmt.Errorf("replaying event[%d] in PCR[%d]: %v", i, pcr.Index, err)
 		}
-		replay[e.index] = replayValue
-		events = append(events, event)
+		replay = replayValue
+		outEvents = append(outEvents, Event{sequence: e.sequence, Data: e.data, Digest: digest, Index: pcr.Index, Type: e.typ})
+	}
+
+	if hadEvents && !bytes.Equal(replay, pcr.Digest) {
+		return nil, fmt.Errorf("PCR[%d] failed to replay using crypto.Hash(%v)", pcr.Index, pcr.DigestAlg)
```
ericchiang: nit: HashAlg already has a Stringer implementation

attest/eventlog.go
```diff
@@ -241,46 +249,88 @@ func extend(pcr, replay []byte, e rawEvent) ([]byte, Event, error) {
...
-		pcrByIndex[pcr.Index] = pcr.Digest
-	}
+// replayPCR replays the event log for a specific PCR, using pcr and
+// event digests with the algorithm in pcr. An error is returned if the
+// replayed values do not match the final PCR digest, or any event tagged
+// with that PCR does not posess an event digest with the specified algorithm.
+func replayPCR(rawEvents []rawEvent, pcr PCR) ([]Event, error) {
+	var (
+		replay    []byte
+		hadEvents bool
```
ericchiang: nit: instead of `hadEvents`, just do `len(outEvents) > 0`

attest/eventlog.go
```diff
@@ -241,46 +249,88 @@ func extend(pcr, replay []byte, e rawEvent) ([]byte, Event, error) {
...
-	replay := map[int][]byte{}
-	pcrByIndex := map[int][]byte{}
-	for _, pcr := range pcrs {
-		pcrByIndex[pcr.Index] = pcr.Digest
-	}
+// replayPCR replays the event log for a specific PCR, using pcr and
+// event digests with the algorithm in pcr. An error is returned if the
+// replayed values do not match the final PCR digest, or any event tagged
+// with that PCR does not posess an event digest with the specified algorithm.
+func replayPCR(rawEvents []rawEvent, pcr PCR) ([]Event, error) {
```
ericchiang: Since replayPCR's error is never reported, can the return value be `[]Event, bool`?

attest/eventlog.go
```diff
@@ -241,46 +249,88 @@ func extend(pcr, replay []byte, e rawEvent) ([]byte, Event, error) {
...
+		allPCRReplays  = map[int][]pcrReplayResult{}
+	)
 
-	var invalidReplays []int
-	for i, value := range replay {
-		if !bytes.Equal(value, pcrByIndex[i]) {
-			invalidReplays = append(invalidReplays, i)
+	// Replay the event log for every PCR and digest algorithm combination.
+	for _, pcr := range pcrs {
+		events, replayErr := replayPCR(rawEvents, pcr)
```
ericchiang: instead of an error, maybe

```
events, ok := replayPCR(rawEvents, pcr)
if ok {
    allPCRReplays[pcr.Index] = append(allPCRReplays[pcr.Index], pcrReplayResult{events})
}
```

attest/eventlog.go
```diff
@@ -224,10 +232,10 @@ var hashBySize = map[int]crypto.Hash{
 	crypto.SHA256.Size(): crypto.SHA256,
 }
 
-func extend(pcr, replay []byte, e rawEvent) ([]byte, Event, error) {
+func extend(pcr, replay []byte, e rawEvent) ([]byte, []byte, error) {
```
ericchiang: since this is returning two values of the same type, they need to be named

```
(extension, digest []byte, err error)
```

attest/eventlog.go
```diff
@@ -186,6 +191,9 @@ func (a *AIKPublic) validate20Quote(quote Quote, pcrs []PCR, nonce []byte) error
 	if err != nil {
 		return fmt.Errorf("parsing quote signature: %v", err)
 	}
+	if att.Magic != tpm20GeneratedMagic {
```
ericchiang: nit: separate PR?

attest/eventlog.go
twitchy-jsonp: `DigestAlg` is `crypto.Hash`, not `HashAlg`. We can move it to use HashAlg if you want?

attest/eventlog.go
twitchy-jsonp: Done.

attest/eventlog.go
ericchiang: nah, it's fine as is. plus we just agreed to return a bool instead of an error anyway.

attest/eventlog.go
twitchy-jsonp: In that case, `allPCRReplays` would only contain successful replays, so while we would omit returning events from unverified PCRs we would not error if both failed to replay?

attest/eventlog.go
twitchy-jsonp: Done.

attest/eventlog.go
twitchy-jsonp: Okies

PR #117, opened 2019-10-01
First attempt at adding support for attribute certificates

mjg59:
Platform certificates are defined as RFC5755 attribute certificates with
various additional attributes and extensions defined in the TCG Platform
Certificate Profile. Add support for parsing them, derived from
crypto/x509.


attributecert/attributecert.go
```diff
@@ -0,0 +1,713 @@
...
+	out.Holder.Serial = in.TBSAttributeCertificate.Holder.BaseCertificateID.Serial
+
+	out.NotBefore = in.TBSAttributeCertificate.Validity.NotBefore
+	out.NotAfter = in.TBSAttributeCertificate.Validity.NotAfter
+
+	var attributes []attribute
+	for _, rawAttr := range in.TBSAttributeCertificate.RawAttributes {
+		var parsedAttribute attribute
+		rest, err := asn1.Unmarshal(rawAttr.FullBytes, &parsedAttribute)
+		if err == nil && len(rest) == 0 {
```
twitchy-jsonp: Nit: handle the error case before the regular case (short-circuit pattern):

```go
switch {
  case err != nil:
    return nil, err
  case len(rest) > 0:
    return nil, errors.New("etc")
}
attributes = append(attributes, parsedAttribute)
```

attributecert/attributecert.go
```diff
@@ -0,0 +1,713 @@
...
+	_, err = asn1.Unmarshal(in.TBSAttributeCertificate.Holder.BaseCertificateID.Issuer.Bytes, &v)
+	if err != nil {
+		return nil, err
+	}
+
+	var holder pkix.RDNSequence
+	if rest, err := asn1.Unmarshal(v.Bytes, &holder); err != nil {
+		return nil, err
+	} else if len(rest) != 0 {
+		return nil, errors.New("x509: trailing data after X.509 subject")
```
twitchy-jsonp: Use the package name as the error prefix:

`attributecert: trailing data after X.509 subject"`

Ditto elsewhere

attributecert/attributecert.go
```diff
@@ -0,0 +1,713 @@
...
+	out.Holder.Issuer.FillFromRDNSequence(&holder)
+	out.Holder.Serial = in.TBSAttributeCertificate.Holder.BaseCertificateID.Serial
+
+	out.NotBefore = in.TBSAttributeCertificate.Validity.NotBefore
+	out.NotAfter = in.TBSAttributeCertificate.Validity.NotAfter
+
+	var attributes []attribute
+	for _, rawAttr := range in.TBSAttributeCertificate.RawAttributes {
+		var parsedAttribute attribute
+		rest, err := asn1.Unmarshal(rawAttr.FullBytes, &parsedAttribute)
```
twitchy-jsonp: Could `RawAttributes` be of type `attribute` rather than `RawValue` ? Would save us parsing it separately.

attributecert/attributecert.go
```diff
@@ -0,0 +1,713 @@
...
+		if err == nil && len(rest) == 0 {
+			attributes = append(attributes, parsedAttribute)
+		} else {
+			return nil, err
+		}
+	}
+
+	for _, attribute := range attributes {
+		if attribute.Id.Equal(oidAttributeUserNotice) {
```
twitchy-jsonp: Use a switch statement for this cascade:

```go
switch {
 case attribute.Id.Equal(oidAttributeUserNotice):
   // ...
  default:
    // ...
   ...
}

File: attributecert/attributecert.go [hunk 5]
```diff
@@ -0,0 +1,713 @@
...
+			for _, value := range attribute.RawValues {
```
twitchy-jsonp: If this is a scalar field, whats the intent of using a for loop over the `RawValues` field, if only the last one will win? what am I missing?

attributecert/attributecert.go
```diff
@@ -0,0 +1,713 @@
...
+				return nil, err
+			}
+		} else if attribute.Id.Equal(oidTcgPlatformConfigUri) {
+			var platformConfigurationUri UriReference
+			_, err := asn1.Unmarshal(attribute.RawValues[0].FullBytes, &platformConfigurationUri)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			return nil, fmt.Errorf("unknown attribute %v", attribute.Id)
```
twitchy-jsonp: maybe add the package prefix to the error for consistency?

Ditto below.

attributecert/attributecert.go
```diff
@@ -0,0 +1,713 @@
...
+				return nil, errors.New("x509: trailing data after X.509 extension")
+			}
+			rest = seq.Bytes
+			for len(rest) > 0 {
+				var v asn1.RawValue
+				rest, err = asn1.Unmarshal(rest, &v)
+				if err != nil {
+					return nil, err
+				}
+				if v.Tag == asn1.TagSet {
```
twitchy-jsonp: The logic for pulling out the different fields seems like a good candidate to decompose into its own method, maybe an unexported `populateFromTCGData()` method?

attributecert/attributecert.go
mjg59: Yes, that seems to work.

attributecert/attributecert.go
mjg59: It's defined as a set of values, but given we're not using that the loop is unnecessary.

mjg59:
Ok, should be in a better state now.

attributecert/attributecert.go
```diff
@@ -0,0 +1,712 @@
...
+	ComponentIdentifiers  []ComponentIdentifierV1 `asn1:"optional,tag=0"`
+	PlatformProperties    []Property              `asn1:"optional,tag=1"`
+	PlatformPropertiesUri []UriReference          `asn1:"optional,tag=2"`
+}
+
+func unmarshalSAN(v asn1.RawValue) ([]pkix.AttributeTypeAndValue, error) {
+	var attributes []pkix.AttributeTypeAndValue
+	if v.Tag == asn1.TagSet {
+		var e pkix.AttributeTypeAndValue
+		_, err := asn1.Unmarshal(v.Bytes, &e)
```
twitchy-jsonp: Nit: This can be rolled into one line. (`if _, err := ...; err != nil { ...`)

attributecert/attributecert.go
```diff
@@ -0,0 +1,712 @@
...
+		if err != nil {
+			return nil, err
+		}
+		attributes = append(attributes, e)
+		return attributes, nil
```
twitchy-jsonp: Instead of appending to an always-empty slice, could we just return this as a literal value?

```go
return []pkix.AttributeTypeAndValue{e}, nil
```

attributecert/attributecert.go
```diff
@@ -0,0 +1,712 @@
...
+	} else if v.Tag == asn1.TagOctetString {
+		var platformData PlatformDataSequence
+		rest, err := asn1.Unmarshal(v.Bytes, &platformData)
+		if err != nil {
+			return nil, err
+		} else if len(rest) != 0 {
+			return nil, errors.New("attributecert: trailing data after X.509 subject")
+		}
+		for _, e := range platformData {
```
twitchy-jsonp: Please bring the declaration of the `attributes` local to here, so its clear its not relevant to the other branch of the if clause.

attributecert/attributecert.go
```diff
@@ -0,0 +1,712 @@
...
+			for _, e2 := range e {
+				attributes = append(attributes, e2)
+			}
+		}
+	}
+	return attributes, nil
```
twitchy-jsonp: This can be `return nil, nil`

attributecert/attributecert.go
```diff
@@ -0,0 +1,712 @@
...
+		if err != nil {
+			return nil, err
+		} else if len(rest) != 0 {
+			return nil, errors.New("attributecert: trailing data after X.509 subject")
+		}
+		for _, e := range platformData {
+			for _, e2 := range e {
+				attributes = append(attributes, e2)
+			}
+		}
```
twitchy-jsonp: For consistency with the other branch, do a `return attributes, nil` here, so the return at the close of the function only needs to handle the case where neither branch is hit.

attributecert/attributecert.go
```diff
@@ -0,0 +1,712 @@
...
+
+func parseAttributeCertificate(in *attributeCertificate) (*AttributeCertificate, error) {
+	out := new(AttributeCertificate)
+	out.Raw = in.Raw
+	out.RawTBSAttributeCertificate = in.TBSAttributeCertificate.Raw
+
+	out.Signature = in.SignatureValue.RightAlign()
+	out.SignatureAlgorithm = getSignatureAlgorithmFromAI(in.TBSAttributeCertificate.SignatureAlgorithm)
+	out.Version = in.TBSAttributeCertificate.Version + 1
+	out.SerialNumber = in.TBSAttributeCertificate.SerialNumber
```
twitchy-jsonp: Nit: This might be cleaner as:

```go
out := &AttributeCertificate{
  Raw: in.Raw,
  RawTBSAttributeCertificate: in.RawTBSAttributeCertificate.Raw,
  Signature: in.SignatureValue.RightAlign(),
  SignatureAlgorithm: getSignatureAlgorithmFromAI(in.TBSAttributeCertificate.SignatureAlgorithm),
  Version: in.TBSAttributeCertificate.Version + 1,
  SerialNumber: in.TBSAttributeCertificate.SerialNumber,
}
```

attributecert/attributecert.go
```diff
@@ -0,0 +1,712 @@
...
+
+	var v asn1.RawValue
+	_, err := asn1.Unmarshal(in.TBSAttributeCertificate.Issuer.IssuerName.Bytes, &v)
+	if err != nil {
+		return nil, err
+	}
```
twitchy-jsonp: Nit: Consider rolling these into one line, here and below, ie:

```go
if _, err := asn1.Unmarshal(in.TBSAttributeCertificate.Issuer.IssuerName.Bytes, &v); err != nil {
	return nil, err
}
```

attributecert/attributecert.go
```diff
@@ -0,0 +1,712 @@
...
+			if err != nil {
+				return nil, err
+			}
+		default:
+			return nil, fmt.Errorf("attributecert: unknown attribute %v", attribute.Id)
+		}
+	}
+
+	for _, extension := range in.TBSAttributeCertificate.Extensions {
+		switch {
```
twitchy-jsonp: Nit: Please add a newline between clauses of the outmost switch block, so its a little easier to visually scan through.

PR #118, opened 2019-10-03
Add config struct for AttestPlatform(), to configure event log source

twitchy-jsonp:
There are two main use-cases for this:

 - Pre-processing of the event log, such as combining the BIOS event log with a log of events extended post-boot
 - Allowing `AttestPlatform()` to succeed even if the system does not have an event log.


twitchy-jsonp:
As discussed IRL, I've:

 - Added an example to demonstrate use of `AttestPlatform()`, and included it with test coverage
 - Updated doc strings to push consumers to use AttestPlatform instead of calling lower-level APIs.

Issue #102, opened 2019-09-19
attest: tests and package should build on non windows/linux+cgo platforms


ericchiang:
There's no reason you shouldn't be able to run the platform independent code on Mac, or on a Linux server without cgo.

```
$ GOOS=darwin go test -v -c github.com/google/go-attestation/attest
# github.com/google/go-attestation/attest [github.com/google/go-attestation/attest.test]
./attest.go:88:9: undefined: platformTPM
./attest.go:90:26: undefined: platformTPM
./attest.go:91:11: undefined: platformTPM
./tpm.go:274:7: undefined: platformTPM
$ GOOS=linux CGO_ENABLED=0 go test -c github.com/google/go-attestation/attest
build github.com/google/go-tspi/tspi: cannot load github.com/google/go-tspi/tspi: no Go source files
```

josephlr:
This seems fixed now. Running
```bash
GOOS=darwin go test -v -c github.com/google/go-attestation/attest
GOOS=darwin go build ./...
```
works fine.

PR #119, opened 2019-10-03
Add Dockerfile for local testing

brandonweeks:


ericchiang:
Would https://github.com/google/go-attestation/issues/102 be a better solution here?

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+RUN apt-get update && apt-get install -y \
+    libtspi-dev \
+ && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /go/src/app
+COPY . .
+
+RUN go test -i -v ./...
+
+CMD cd attest && go test -v
```
ericchiang: RUN and CMD are redundant here, right? You could just use one of them

Issue #126, opened 2019-10-08
parseRawEvent can allocate an unbounded amount of memory


brandonweeks:
```go
var h rawEventHeader
if err = binary.Read(r, binary.LittleEndian, &h); err != nil {
    return event, err
}
data := make([]byte, int(h.EventSize))
```
In this code path an attacker can control the number of allocated bytes. This can lead to a DoS attack by OOMing the process.

Example:
```
Mzk0MDIwMDYxOTYzOTQ0NzkyMTIyNzkwNDAxbUfvv70AMDAxNDM2MTM4MDUwNzk3MzkyNzA0NjU0
NDY2Njc5NDgyOTM0MDQyNDU3MjE3NzE0OTY4NzAzMjkwNDcyNjYwODgyNTg5MzgwMDE4NjE2MDY5
NzMxMTIzMTk=
```
Produces:
```go
rawEventHeader{
    PCRIndex:0x30343933, 
    Type:0x36303032, 
    Digest:[20]uint8{0x31, 0x39, 0x36, 0x33, 0x39, 0x34, 0x34, 0x37, 0x39, 0x32, 0x31, 0x32, 0x32, 0x37, 0x39, 0x30, 0x34, 0x30, 0x31, 0x6d},
    EventSize:0xbdbfef47,
}
```
`0xbdbfef47` being 3.183 GB.

It doesn't appear that the [TCG EFI Protocol Specification](https://trustedcomputinggroup.org/wp-content/uploads/EFI-Protocol-Specification-rev13-160330final.pdf) defines a maximum size for an event. So it seems our options are either choosing an arbitrary maximum or reporting this to the TCG as undefined behavior.

ericchiang:
Nice catch :) 

What about adding a check to ensure h.EventSize isn't bigger than the remaining bytes in the measurement log?

```go
func parseRawEvent(r *bytes.Buffer) (event rawEvent, err error) {
	var h rawEventHeader
	if err = binary.Read(r, binary.LittleEndian, &h); err != nil {
		return event, err
	}
	if h.EventSize > uint32(r.Len()) {
		return event, fmt.Errorf("event size (%d) was greater than remaining bytes in measurement log (%d)", h.EventSize, r.Len())
	}
	data := make([]byte, int(h.EventSize))
	if _, err := io.ReadFull(r, data); err != nil {
		return event, err
	}
	return rawEvent{
		typ:     EventType(h.Type),
		data:    data,
		index:   int(h.PCRIndex),
[...]

PR #127, opened 2019-10-08
attest: ensure parsing event can't allocated unbounded memory

ericchiang:
Add a sanity check to ensure the measurement log actually contains as
much data as the event size reports.

Fixes https://github.com/google/go-attestation/issues/126


ericchiang:
Actually, let me add a test

ericchiang:
Test added. Now with more Go 1.13.

Issue #132, opened 2019-10-11
Rewire HashAlg.cryptoHash() once go-tpm#143 lands


twitchy-jsonp:
The PR pending here: https://github.com/google/go-tpm/pull/143

Will export a method to get a `crypto.Hash` from a `tpm2.Algorithm`. We should use it once it lands so we have a smaller integration surface to maintain.

PR #133, opened 2019-10-17
attest: work around tpm2 removal of Algorithm.HashConstructor()

ericchiang:
go-tpm removed HashConstructor in favor of Hash breaking our build at
tip. To ensure go-attestation works with as many versions of go-tpm as
possible, switch to our internal method for converting a tpm2.Algorithm
to a crypto.Hash (instead of using the new Algorithm.Hash()).

https://github.com/google/go-tpm/commit/431489f432542ec0b7acd4a5c9b8dbbde56f2869

### DIFF
--- a/attest/activation.go
+++ b/attest/activation.go
@@ -121,11 +121,11 @@ func (p *ActivationParameters) checkTPM20AKParameters() error {
 
 	// Compute & verify that the creation data matches the digest in the
 	// attestation structure.
-	nameHashConstructor, err := pub.NameAlg.HashConstructor()
+	nameHash, err := pub.NameAlg.Hash()
 	if err != nil {
 		return fmt.Errorf("HashConstructor() failed: %v", err)
 	}
-	h := nameHashConstructor()
+	h := nameHash.New()
 	h.Write(p.AK.CreateData)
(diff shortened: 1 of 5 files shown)


twitchy-jsonp:
Hmmm, wheres our CI thingy that tests against head (as opposed to the version in go.mod?)

ericchiang:
What did we just enable? If I'm not up to date with master I can't merge, and re-basing locally and force pushing dismisses your approval...

ericchiang:
> Hmmm, wheres our CI thingy that tests against head (as opposed to the version in go.mod?)

I think there change was more recent than our lastest PR. So since we haven't run CI, we haven't caught it?

josephlr:
This PR is a little confusing. Why not just update go.mod and use the new methods in go-tpm? Isn’t that the whole reason go.mod exists?

ericchiang:
@josephlr fair enough. Updated.

Does go-tpm expect to make more breaking API changes? The google monorepo doesn't use go.mod, but as long as you're okay with updating go-tpm and go-attestation at the same time that works for me :)

twitchy-jsonp:
Fixes #132

PR #138, opened 2019-11-13
attest: add code for turning the AK into a signer

ericchiang:
The added test currently fails with the following error message:

    === RUN   TestSimTPM20RSASign
    --- FAIL: TestSimTPM20RSASign (0.10s)
        attest_simulated_tpm20_test.go:291: signing failed: signing with scheme &tpm2.SigScheme{Alg:0x14, Hash:0xb, Count:0x0} failed: parameter 3, error code 0x20 : invalid ticket

"invalid ticket" would appear to imply that the hash algorithm isn't the
one specified in the key scheme[1]. Overall I'm kinda confused.

[1] https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-3-Commands-01.38-code.pdf#page=204

### DIFF
--- a/attest/signer.go
+++ b/attest/signer.go
@@ -1,0 +1,128 @@
+package attest
+
+import (
+	"crypto"
+	"crypto/rsa"
+	"encoding/asn1"
+	"fmt"
+	"io"
+	"math/big"
+
+	"github.com/google/go-tpm/tpm2"
+	"github.com/google/go-tpm/tpmutil"
+)
+
+var tpm2HashFuncMap = map[crypto.Hash]tpm2.Algorithm{
+	crypto.SHA1:   tpm2.AlgSHA1,
+	crypto.SHA256: tpm2.AlgSHA256,
+	crypto.SHA384: tpm2.AlgSHA384,
+	crypto.SHA512: tpm2.AlgSHA512,
+}
+
+func tpm2HashFunc(h crypto.Hash) (tpm2.Algorithm, error) {
+	if a, ok := tpm2HashFuncMap[h]; ok {
+		return a, nil
+	}
+	return 0, fmt.Errorf("unsupported hash algorithm 0x%x", h)
+}
+
+type ec20Key struct {
+	tpm    io.ReadWriter
+	handle tpmutil.Handle
+	pub    crypto.PublicKey
+}
+
+type ecdsaSignature struct {
+	R, S *big.Int
+}
+
+func (k *ec20Key) Public() crypto.PublicKey {
+	return k.pub
+}
+
+func (k *ec20Key) Sign(rand io.Reader, digest []byte, opts crypto.SignerOpts) ([]byte, error) {
+	hash, err := tpm2HashFunc(opts.HashFunc())
+	if err != nil {
+		return nil, err
+	}
+	scheme := &tpm2.SigScheme{Hash: hash, Alg: tpm2.AlgECDSA}
+	sig, err := tpm2.Sign(k.tpm, k.handle, akDefaultOwnerPassword, digest, scheme)
+	if err != nil {
+		return nil, fmt.Errorf("signing failed: %v", err)
+	}
+	if sig.ECC == nil {
+		return nil, fmt.Errorf("signing operation didn't return a signature")
+	}
+	// See: https://golang.org/src/crypto/ecdsa/ecdsa.go
+	return asn1.Marshal(ecdsaSignature{sig.ECC.R, sig.ECC.S})
+}
+
+type rsa20Key struct {
+	tpm    io.ReadWriter
+	handle tpmutil.Handle
+	pub    crypto.PublicKey
+}
+
+func (k *rsa20Key) Public() crypto.PublicKey {
+	return k.pub
+}
+
+type sigErr struct {
+	scheme *tpm2.SigScheme
+	err    error
+}
+
+func (s *sigErr) Error() string {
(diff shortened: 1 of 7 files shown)


attest/signer.go
```diff
@@ -0,0 +1,128 @@
...
+func (k *ec20Key) Public() crypto.PublicKey {
+	return k.pub
+}
+
+func (k *ec20Key) Sign(rand io.Reader, digest []byte, opts crypto.SignerOpts) ([]byte, error) {
+	hash, err := tpm2HashFunc(opts.HashFunc())
+	if err != nil {
+		return nil, err
+	}
+	scheme := &tpm2.SigScheme{Hash: hash, Alg: tpm2.AlgECDSA}
```
twitchy-jsonp: TPM 2.0 Commands reference section 20.2.1 says:

```
If the scheme of keyHandle is not TPM_ALG_NULL, then inScheme shall either be the same scheme as
keyHandle or TPM_ALG_NULL. If the sign attribute is not SET in the key referenced by handle then the
TPM shall return TPM_RC_KEY
```

For most AKs (including ours) the Scheme is RSASSA with a SHA256 hash: https://github.com/google/go-attestation/blob/master/attest/tpm.go#L53, so for those keys a NULL scheme or an identical scheme will need to be provided.

The AK template has `tpm2.FlagSignerDefault` set, which includes the sign bit, so we should be fine with regards to key permissions.

attest/signer.go
```diff
@@ -0,0 +1,128 @@
...
+	sig, err := tpm2.Sign(k.tpm, k.handle, akDefaultOwnerPassword, digest, scheme)
```
twitchy-jsonp: TPM 2.0 Commands reference section 20.2.1 says:

```
If keyHandle references a restricted signing key, then validation shall be provided, indicating that the TPM
performed the hash of the data and validation shall indicate that hashed data did not start with
TPM_GENERATED_VALUE
```

The `tpm2.Sign` / `tpm2.SignWithSession` function [always uses a NULL ticket](https://github.com/google/go-tpm/blob/f0607eac7f8a894e8a26b848c072626f1cfe8b2d/tpm2/tpm2.go#L1328) with tag `TPM_ST_CHECKHASH` for the _validation_ argument, so the function in its current form will never work for AKs.

attest/signer.go
```diff
@@ -0,0 +1,128 @@
...
+type sigErr struct {
+	scheme *tpm2.SigScheme
+	err    error
+}
+
+func (s *sigErr) Error() string {
+	return fmt.Sprintf("signing with scheme %#v failed: %v", s.scheme, s.err)
+}
+
+func (k *rsa20Key) Sign(rand io.Reader, digest []byte, opts crypto.SignerOpts) ([]byte, error) {
```
twitchy-jsonp: Ditto above.

ericchiang:
> AKs can only sign using the algorithms specified in their template (unless NULL is specified in their template, which usually isnt the case).

Is there a strong reason not to specify NULL? It seems like a good thing for the AK not to be constrained to a single algorithm.

> Calling TPM2_Sign on an AK requires providing a proof (in the form of a ticket) that the data does not start with TPM_GENERATED_VALUE. Providing such a ticket is not currently implemented in go-tpm.

Ah that's the issue here. Any reason not to add that code to go-tpm? Are there alternative ways to perform signing?

There were also some offline discussions about using AKs for signing vs. using EKs for signing vs. generating another key for signing. Based on the example in the README, it seems more straightforward for the EK or AK to do the signing since that's what's being attested in the identity proof. Does that logic follow or am I missing other advantages for having a third key?

twitchy-jsonp:
> Is there a strong reason not to specify NULL? It seems like a good thing for the AK not to be constrained to a single algorithm.

On Windows, we don't control the template the AK is created with - Its hardcoded to RSASSA / SHA256.

This field also controls the scheme used for quotes + key certification. Code which assumes this is set (ie: code that calls Certify/CertifyCreation/Quote/Quote2 with a null signing scheme) will break.

> Any reason not to add that code to go-tpm? Are there alternative ways to perform signing?

Nope! feel free :)

The alternative ways to get an AK to sign something involve having the TPM generate the data to be signed (ie: quote/certify)

> Based on the example in the README, it seems more straightforward for the EK or AK to do the signing since that's what's being attested in the identity proof. Does that logic follow or am I missing other advantages for having a third key?

From a user-story standpoint its certainly simpler if there was only the 'AK' and it could be used for signing (The EK can only be used for signing if it was generated with that bit set, which is seldom the case).
However theres heaps of edge-cases here (the scheme, the TPMS_GENERATED_VALUE gotcha etc) that mean it might be hard for a user to reason about using an AK for signing, or make it inpractical.
As such, if we want to support arbitrary signatures in go-attestation, I'm a proponent of creating a new package-level type to represent a signing key. Keeping AK's and signing keys separate also maps onto the model for TPM 1.2, so we could support TPM 1.2 as well.

attest/signer.go
tomoveu: @twitchy-jsonp great explanation. I would suggest @ericchiang to create new API tpm2.SignWithTicket instead of modifying or extending the existing ones. Just my 2 cents.

ericchiang: This was taken over by https://github.com/google/go-attestation/pull/201 :) which does the right thing to create a crypto.Signer compatible interface

ericchiang:
Closing in favor of https://github.com/google/go-attestation/pull/201

PR #140, opened 2019-12-04
Validate signature scheme is present when decoding TPMT_PUBLIC blobs

### LINKED ISSUE #139: Segfault in ParseAKPublic
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x2 pc=0x56251519dc69]

<redacted>/attest.ParseAKPublic(0x2, 0xc0029c44e0, 0x1a, 0x20, 0xc002a13980, 0x1, 0x1)
        <redacted>/attest/attest.go:244 +0x449 fp=0xc002a63090 sp=0xc002a62fa8 pc=0x56251519dc69
```

Appears to be a deref somewhere in `.RSAParameters.Sign.Hash`.

This input can cause it: `[0 1 0 4 0 1 0 0 0 0 0 6 0 128 0 67 0 16 8 0 0 1 0 1 0 0]`

twitchy-jsonp:
Fixes #139

### DIFF
--- a/attest/attest.go
+++ b/attest/attest.go
@@ -234,10 +234,20 @@ func ParseAKPublic(version TPMVersion, public []byte) (*AKPublic, error) {
 		if err != nil {
 			return nil, fmt.Errorf("parsing TPM public key structure: %v", err)
 		}
+		switch {
+		case pub.RSAParameters == nil && pub.ECCParameters == nil:
+			return nil, errors.New("parsing public key: missing asymmetric parameters")
+		case pub.RSAParameters != nil && pub.RSAParameters.Sign == nil:
+			return nil, errors.New("parsing public key: missing rsa signature scheme")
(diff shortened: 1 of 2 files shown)


attest/attest.go
```diff
@@ -234,10 +234,14 @@ func ParseAKPublic(version TPMVersion, public []byte) (*AKPublic, error) {
 		if err != nil {
 			return nil, fmt.Errorf("parsing TPM public key structure: %v", err)
 		}
+		if pub.RSAParameters.Sign == nil {
```
ericchiang: One question, what if this is a EC key?

ericchiang: Would the check be better in the switch statement below?

twitchy-jsonp: Added validation for both RSA & ECC parameters.

PR #141, opened 2019-12-04
Enable ParseAKPublic & ParseEKCertificate for fuzzing.

twitchy-jsonp:
The other major method is `ActivateParameters{}.Generate()`, but I'll leave that for another PR (as we need to generate corpus and a bunch of stuff to get that working meaningfully).


ericchiang:
nit: can this be in a file called *_fuzz.go? Maybe attest_fuzz.go

twitchy-jsonp:
Done :)

PR #144, opened 2019-12-10
Pay attention to digest size information in the headers

mjg59:
Crypto agile logs may contain digest types that we don't currently
handle. However, we still need to know how long each digest is in order
to read over the appropriate amount of the buffer. This information is
provided to us as part of the spec header - make use of it rather than
hardcoding the set of digests and lengths we know about.

### DIFF
--- a/attest/eventlog.go
+++ b/attest/eventlog.go
@@ -343,20 +343,21 @@ const eventTypeNoAction = 0x03
 
 // ParseEventLog parses an unverified measurement log.
 func ParseEventLog(measurementLog []byte) (*EventLog, error) {
+	var specID *specIDEvent
 	r := bytes.NewBuffer(measurementLog)
 	parseFn := parseRawEvent
 	var el EventLog
-	e, err := parseFn(r)
+	e, err := parseFn(r, specID)
 	if err != nil {
 		return nil, fmt.Errorf("parse first event: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


attest/eventlog.go
```diff
@@ -529,20 +530,20 @@ func parseRawEvent2(r *bytes.Buffer) (event rawEvent, err error) {
...
-		case algSHA1:
-			digest = make([]byte, crypto.SHA1.Size())
-		case algSHA256:
-			digest = make([]byte, crypto.SHA256.Size())
-		default:
-			// ignore signatures that aren't SHA1 or SHA256
-			continue
+		for _, alg := range specID.algs {
+			if alg.ID == algID {
+				digest = make([]byte, alg.Size)
```
ericchiang: nit: add a check here to make sure this can't allocate a huge amount of memory and add a break

```
if alg.ID != algID {
    continue
}
if len(r.Len()) > alg.Size {
    return event, fmt.Errorf("reading digest: %v", io.ErrUnexpectedEOF)
}
digest = make([]byte, alg.Size)
break
```

mjg59:
```
if len(r.Len()) > alg.Size {
```
Do you mean the other way around here?

ericchiang: woops, yeah you're right

PR #148, opened 2019-12-17
Implement extractor for determining secure boot state

twitchy-jsonp:


attest/internal/events.go
```diff
@@ -0,0 +1,329 @@
...
+		return ""
+	}
+	if err := binary.Write(&buf, binary.BigEndian, d.Data3); err != nil {
+		return ""
+	}
+	if err := binary.Write(&buf, binary.BigEndian, d.Data4); err != nil {
+		return ""
+	}
+	uuid := uuid.UUID{}
+	if err := uuid.UnmarshalBinary(buf.Bytes()); err != nil {
```
ericchiang: These's almost no logic being done here:

https://github.com/google/uuid/blob/c2e93f3ae59f2904160ceaab466009f965df46d6/marshal.go#L31-L36
https://github.com/google/uuid/blob/c2e93f3ae59f2904160ceaab466009f965df46d6/uuid.go#L164

Can we skip importing github.com/google/uuid and just do:

```go
func (d efiGUID) String() string {
	var u [8]byte
	binary.BigEndian.PutUint32(u[:4], d.Data1)
	binary.BigEndian.PutUint16(u[4:6], d.Data2)
	binary.BigEndian.PutUint16(u[6:8], d.Data3)
	return fmt.Sprintf("%x-%x-%x-%x-%x", u[:4], u[4:6], u[6:8], d.Data4[:2], d.Data4[2:])
}
```

attest/internal/events.go
```diff
@@ -0,0 +1,329 @@
...
+}
+
+// ParseUEFIVariableData parses the data section of an event structured as
+// a UEFI variable.
+func ParseUEFIVariableData(r io.Reader) (ret UEFIVariableData, err error) {
+	err = binary.Read(r, binary.LittleEndian, &ret.Header)
+	if err != nil {
+		return
+	}
+	ret.UnicodeName = make([]uint16, ret.Header.UnicodeNameLength)
```
ericchiang: Need to check `ret.Header.UnicodeNameLength` for malicious values?

attest/internal/events.go
```diff
@@ -0,0 +1,329 @@
...
+	for i := 0; uint64(i) < ret.Header.UnicodeNameLength; i++ {
+		err = binary.Read(r, binary.LittleEndian, &ret.UnicodeName[i])
+		if err != nil {
+			return
+		}
+	}
+	ret.VariableData = make([]byte, ret.Header.VariableDataLength)
```
ericchiang: dito for `VariableDataLength`

attest/internal/events.go
```diff
@@ -0,0 +1,329 @@
...
+
+// UEFIVariableAuthority describes the contents of a UEFI variable authority
+// event.
+type UEFIVariableAuthority struct {
+	Certs []x509.Certificate
+}
+
+// ParseUEFIVariableAuthority parses the data section of an event structured as
+// a UEFI variable authority.
+func ParseUEFIVariableAuthority(r io.Reader) (UEFIVariableAuthority, error) {
```
ericchiang: Can one of these parse functions link to the spec?

(tip, you can add #page=N to link to a page in a PDF)

attest/internal/events.go
```diff
@@ -0,0 +1,329 @@
...
+		err := binary.Read(buf, binary.LittleEndian, &signatures.Header)
+		if err != nil {
+			return nil, nil, err
+		}
+		signatureType := signatures.Header.SignatureType.String()
+		switch signatureType {
+		case "a5c059a1-94e4-4aa7-87b5-ab155c2bf072": // X509 certificate
+			for sigOffset := 0; uint32(sigOffset) < signatures.Header.SignatureListSize-28; {
+				signature := efiSignatureData{}
+				signature.SignatureData = make([]byte, signatures.Header.SignatureSize-16)
```
ericchiang: dito: need to check signatures.Header.SignatureSize

attest/internal/events.go
```diff
@@ -0,0 +1,329 @@
...
+	hashes := [][]byte{}
+
+	for offset := 0; offset < len(b); {
+		err := binary.Read(buf, binary.LittleEndian, &signatures.Header)
+		if err != nil {
+			return nil, nil, err
+		}
+		signatureType := signatures.Header.SignatureType.String()
+		switch signatureType {
+		case "a5c059a1-94e4-4aa7-87b5-ab155c2bf072": // X509 certificate
```
ericchiang: link to something showing where these GUIDs are linked?

Also you can statically define structs with comparable fields and just compare them without using strings

```
var guidX509Certificate = eGUID{
    // raw bytes
}

// ...

switch signature.Header.SignatureType {
case guidX509Certificate:
    // ...
}
```

attest/internal/events.go
```diff
@@ -0,0 +1,329 @@
...
+				if err != nil {
+					return nil, nil, err
+				}
+				sigOffset += int(signatures.Header.SignatureSize)
+				certificates = append(certificates, *cert)
+			}
+		case "c1c41626-504c-4092-aca9-41f936934328": // SHA256
+			for sigOffset := 0; uint32(sigOffset) < signatures.Header.SignatureListSize-28; {
+				signature := efiSignatureData{}
+				signature.SignatureData = make([]byte, signatures.Header.SignatureSize-16)
```
ericchiang: dito for checking SignatureSize

attest/internal/events.go
```diff
@@ -0,0 +1,329 @@
...
+	if len(b) < 28 {
+		// Being passed an empty signature list here appears to be valid
+		return nil, nil, nil
+	}
+	signatures := efiSignatureList{}
+	buf := bytes.NewReader(b)
+	certificates := []x509.Certificate{}
+	hashes := [][]byte{}
+
+	for offset := 0; offset < len(b); {
```
ericchiang: instead of tracking offset, why not use buf.Len?

attest/internal/events.go
```diff
@@ -0,0 +1,329 @@
...
+		err := binary.Read(buf, binary.LittleEndian, &signatures.Header)
+		if err != nil {
+			return nil, nil, err
+		}
+		signatureType := signatures.Header.SignatureType.String()
+		switch signatureType {
+		case "a5c059a1-94e4-4aa7-87b5-ab155c2bf072": // X509 certificate
+			for sigOffset := 0; uint32(sigOffset) < signatures.Header.SignatureListSize-28; {
```
ericchiang: If SignatureListSize was less than 28 would this panic?

Also maybe a limit reader would be better here or some io.Reader way of tracking the offset?

attest/internal/events.go
```diff
@@ -0,0 +1,329 @@
...
+	SignatureOwner efiGUID
+	SignatureData  []byte // []int8
+}
+
+func parseEfiSignature(b []byte) ([]x509.Certificate, error) {
+	certificates := []x509.Certificate{}
+
+	buf := bytes.NewReader(b)
+	signature := EFISignatureData{}
+	signature.SignatureData = make([]byte, len(b)-16)
```
ericchiang: check to make sure len(b) is at least 16?

attest/secureboot.go
```diff
@@ -0,0 +1,166 @@
...
+		if e.Index != 7 {
+			continue
+		}
+
+		et, err := internal.ExtractEventType(uint32(e.Type))
+		if err != nil {
+			return nil, fmt.Errorf("unrecognised event type: %v", err)
+		}
+
+		digestVerify := e.digestEquals(e.Data)
```
ericchiang: You're checking this on everything but EFIAction. Maybe

```
if et != internal.EFIAction {
    if err := e.digestEqual(e.Data); err != nil {
        // return error
    }
}
```

attest/secureboot.go
```diff
@@ -0,0 +1,166 @@
...
+		switch et {
+		case internal.Separator:
+			if seenSeparator {
+				return nil, fmt.Errorf("duplicate separator at event %d", e.sequence)
+			}
+			seenSeparator = true
+			if !bytes.Equal(e.Data, []byte{0, 0, 0, 0}) {
+				return nil, fmt.Errorf("invalid separator data at event %d: %v", e.sequence, e.Data)
+			}
+			if err := e.digestEquals(e.Data); err != nil {
```
ericchiang: You've already computed the digest above, just use that.

attest/secureboot.go
```diff
@@ -0,0 +1,166 @@
...
+			}
+
+		case internal.EFIAction:
+			if string(e.Data) == "UEFI Debug Mode" {
+				return nil, errors.New("a UEFI debugger was present during boot")
+			}
+			return nil, fmt.Errorf("event %d: unexpected EFI action event", e.sequence)
+
+		case internal.EFIVariableDriverConfig:
+			v, err := internal.ParseUEFIVariableData(bytes.NewReader(e.Data))
```
ericchiang: nit, check the digest before parsing?

attest/secureboot.go
```diff
@@ -0,0 +1,166 @@
...
+			} else {
+				out.PostSeparatorAuthority = append(out.PostSeparatorAuthority, a.Certs...)
+			}
+
+		default:
+			return nil, fmt.Errorf("unexpected event type: %v", et)
+		}
+	}
+
+	if out.Enabled {
```
ericchiang: nit:

```
if !out.Enabled {
    return &out, nil
}
// checks
return &out, nil
```

attest/internal/events.go
```diff
@@ -0,0 +1,329 @@
...
+
+func (e EventType) String() string {
+	if s, ok := eventTypeNames[e]; ok {
+		return s
+	}
+	return fmt.Sprintf("EventType(0x%x)", uint32(e))
+}
+
+// ExtractEventType returns the event type indicated by the provided value.
+func ExtractEventType(et uint32) (EventType, error) {
```
ericchiang: nit: can we call this UntrustedExtractEventType? I think it clarifies that this is a dangerous method to use.

attest/secureboot.go
```diff
@@ -0,0 +1,166 @@
...
+	PreSeparatorAuthority  []x509.Certificate
+	PostSeparatorAuthority []x509.Certificate
+}
+
+// ExtractSecurebootState parses a series of events to determine the
+// configuration of secure boot on a device. An error is returned if
+// the state cannot be determined, or if the event log is structured
+// in such a way that it may have been tampered post-execution of
+// platform firmware.
+func ExtractSecurebootState(events []Event) (*SecurebootState, error) {
```
ericchiang: We've got a lot of "Verify" "Parse" etc. synonyms. Instead of adding "Extract" can we just call this "ParseSecurebootState"?

attest/secureboot.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/google/certificate-transparency-go/x509"
+	"github.com/google/go-attestation/attest/internal"
+)
+
+// SecurebootState describes the secure boot status of a machine, as determined
+// by processing its event log.
+type SecurebootState struct {
+	Enabled bool
+
+	PlatformKeys      []x509.Certificate
```
ericchiang: Can you add basic english comments to these exported fields? :)

attest/internal/events.go
twitchy-jsonp: No, theres a guard clause at the beginning of the method.

attest/secureboot.go
twitchy-jsonp: That would make it cleaner, but I prefer having duplication so we get the more specific error messages in this case.

attest/secureboot.go
twitchy-jsonp: Would prefer to parse first so we can have better error messages.

attest/internal/events.go
twitchy-jsonp: I attempted that but couldn't get it working, not sure why.

attest/internal/events.go
```diff
@@ -0,0 +1,369 @@
...
+			}
+		case hashSHA256SigGUID: // SHA256
+			for sigOffset := 0; uint32(sigOffset) < signatures.Header.SignatureListSize-28; {
+				signature := efiSignatureData{}
+				signature.SignatureData = make([]byte, signatures.Header.SignatureSize-16)
+				err := binary.Read(buf, binary.LittleEndian, &signature.SignatureOwner)
+				if err != nil {
+					return nil, nil, err
+				}
+				err = binary.Read(buf, binary.LittleEndian, &signature.SignatureData)
```
ericchiang: this is shadowed by the `err := above`

twitchy-jsonp: Doneskies

PR #151, opened 2020-01-30
Validate a platform event log in AttestPlatform()

### LINKED ISSUE #150: AttestPlatform() returns an empty event log on some systems
This can lead to errors when trying to parse it: `invalid log: event data size is 0`.

Given an empty event log defeats the point of platform attestation, a call to `AttestPlatform()` should return an error if the platform does not have a valid event log.

twitchy-jsonp:
Fixes #150

### DIFF
--- a/attest/tpm.go
+++ b/attest/tpm.go
@@ -302,7 +302,17 @@ func (t *TPM) LoadAK(opaqueBlob []byte) (*AK, error) {
 // This is a low-level API. Consumers seeking to attest the state of the
 // platform should use tpm.AttestPlatform() instead.
 func (t *TPM) MeasurementLog() ([]byte, error) {
-	return t.tpm.measurementLog()
+	el, err := t.tpm.measurementLog()
+	if err != nil {
+		return nil, err
+	}
+
+	// A valid event log contains at least one SpecID event header (28 bytes).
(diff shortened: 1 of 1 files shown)


ericchiang:
Should this check be lower down for better reporting?

https://github.com/google/go-attestation/blob/2bc8d585301fd95b3d119417d27493f6502969e4/attest/tpm_linux.go#L395
https://github.com/google/go-attestation/blob/73020b971b6586b64a09ef6b39106bf391c3c81a/attest/tpm_windows.go#L380

E.g on Windows check the result of `logLen, err := context.GetTCGLog(nil)` and on Linux check `binary_bios_measurements`?

twitchyliquid64:
I had that originally but moved it higher as the logic was identical. All we are doing is checking the returned log is some minimum length, which is a platform agnostic check. I guess we could make the error message platform specific, but also not sure of the utility given the caller should be able to check the current platform, and if a caller ever wanted to inspect the error they would need to handle two error strings.

ericchiang:
Even then, MeasurementLog() should return the error, not only AttestPlatform(), right?

twitchy-jsonp:
Sounds good, I've moved into `MeasurementLog()`

PR #152, opened 2020-02-21
Remove length-based hash lookups

mjg59:
Using the length of a digest to infer the hash algorithm is somewhat
fragile - if we end up with multiple hash algorithms that share the same
digest length, things will break. Instead, pass more complete digest
information through to relevant functions and figure things out by
mapping the TPM hash algorithm to the appropriate Golang type.


attest/eventlog.go
```diff
@@ -273,10 +262,10 @@ func extend(pcr, replay []byte, e rawEvent) (pcrDigest []byte, eventDigest []byt
 			b := make([]byte, h.Size())
 			hash.Write(b)
 		}
-		hash.Write(digest)
-		return hash.Sum(nil), digest, nil
+		hash.Write(digest.data)
+		return hash.Sum(nil), digest.data, nil
 	}
-	return nil, nil, fmt.Errorf("no event digest matches pcr length: %d", len(pcr))
+	return nil, nil, fmt.Errorf("no event digest matches pcr algorithm: %d", pcr.DigestAlg)
```
twitchy-jsonp: Use `%v` as `DigestAlg` is of type `crypto.Hash`

attest/eventlog.go
```diff
@@ -526,15 +520,21 @@ func parseRawEvent(r *bytes.Buffer, specID *specIDEvent) (event rawEvent, err er
 	if h.EventSize > uint32(r.Len()) {
 		return event, &eventSizeErr{h.EventSize, r.Len()}
 	}
+
+	digests := make([]digest, 1)
```
twitchy-jsonp: This can be constructed literally:

```go
digests := []digest{{hash: crypto.SHA1, data: h.Digest[:]}}
```

attest/eventlog.go
```diff
@@ -564,20 +570,25 @@ func parseRawEvent2(r *bytes.Buffer, specID *specIDEvent) (event rawEvent, err e
...
 		for _, alg := range specID.algs {
 			if alg.ID != algID {
 				continue
 			}
 			if uint16(r.Len()) < alg.Size {
 				return event, fmt.Errorf("reading digest: %v", io.ErrUnexpectedEOF)
 			}
-			digest = make([]byte, alg.Size)
+			digest.data = make([]byte, alg.Size)
+			hash, ok := algorithmMap[tpm2.Algorithm(alg.ID)]
```
twitchy-jsonp: I think you can just do `HashAlg(tpm2.Algorithm(alg.ID)).cryptoHash()`, as the values for `attest.HashAlg` and `tpm2.Algorithm` should match

attest/eventlog.go
```diff
@@ -223,47 +223,36 @@ func (a *AKPublic) validate20Quote(quote Quote, pcrs []PCR, nonce []byte) error
...
-	h, ok := hashBySize[len(pcr)]
-	if !ok {
-		return nil, nil, fmt.Errorf("pcr %d was not a known hash size: %d", e.index, len(pcr))
-	}
 	for _, digest := range e.digests {
-		if len(digest) != len(pcr) {
+		if digest.hash != pcr.DigestAlg {
+			continue
+		}
+		if len(digest.data) != len(pcr.Digest) {
```
twitchy-jsonp: Is the digest length check necessary given we have `if digest.hash != pcr.DigestAlg {` above? If yes, could we make it bail with an error instead of `continue` ?

mjg59: Done

PR #153, opened 2020-02-24
Add workaround in validation for missing exit boot services event log messages

twitchy-jsonp:
I removed the logic to try all combinations of workarounds as it wasn't needed yet + its just additional complexity.

Also - I havent set the event type for the injected messages. Is it worth doing so + if so, whats the event type?


twitchy-jsonp:
Friendly ping :)

twitchy-jsonp:
Friendly ping :)

PR #158, opened 2020-04-30
Implement workaround for secureboot authority events from shim

### LINKED ISSUE #157: Secure-boot: parsing authority events starting with `[]byte{185 65 36 160}` fails
Analysis ongoing, but I have a failing unit test:

```
--- FAIL: TestSecureBootBugThing (0.00s)
    secureboot_test.go:58: failed parsing secureboot state: failed parsing EFI variable authority at event 12: asn1: structure error: tags don't match (16 vs {class:2 tag:25 length:65 isCompound:true}) {optional:false explicit:false application:false private:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false lax:true name:} certificate @2
    secureboot_test.go:60: <nil>
```

twitchy-jsonp:
Fixes #157


attest/secureboot.go
```diff
@@ -144,7 +144,13 @@ func ParseSecurebootState(events []Event) (*SecurebootState, error) {
 		case internal.EFIVariableAuthority:
 			a, err := internal.ParseUEFIVariableAuthority(bytes.NewReader(e.Data))
 			if err != nil {
-				return nil, fmt.Errorf("failed parsing EFI variable authority at event %d: %v", e.sequence, err)
+				// Workaround for: https://github.com/google/go-attestation/issues/157
+				if err == internal.ErrSigMissingGUID {
+					// Trim off an erroneous additional byte, and re-do the digest check.
+					digestVerify = e.digestEquals(e.Data[:len(e.Data)-1])
```
ericchiang, line 150: For this case, are you okay with a.Certs being nil? It looks like `ParseUEFIVariableAuthority` throws away the return value from `parseEfiSignature` if the later returns an error.

twitchy-jsonp: Whelp, good catch! ^^' I'll fix + make sure the test checks the cert.

PR #159, opened 2020-05-02
Add CommandChannel to OpenConfig to support downstream simulator

### LINKED ISSUE #156: Make platformTPM public for downstream testing
I am contributing to a downstream project, [spire-tpm-plugin](https://github.com/bloomberg/spire-tpm-plugin), and would like to be able to test using the TPM simulator like this project does:

https://github.com/google/go-attestation/blob/25ce56400c9ff0a90a3f8b3eb47bf1698fe9319a/attest/attest_simulated_tpm20_test.go#L30-L42

The issue I'm running into is that `platformTPM` is a private struct, so I am unable to "fake" the TPM in downstream testing

caleblloyd:
Fixes #156 

Adds `CommandChannel io.ReadWriter` to `OpenConfig` in order to pass network device or simulator to TPM.  `CommandChannel` is currently only supported on TPM2.0 ~~on Linux.  TPM1.2 and TPM2.0 on Windows methods would require significant refactoring to support `io.ReadWriter`.~~

~~Changes Linux TPM2.0 command channel from `io.ReadWriteCloser` to `io.ReadWriter` in order to accommodate streams that do not support `Close()`.  Adds upcast check in `platformTPM.close` method for `io.ReadWriteCloser`.~~

~~Adds `+build linux,!gofuzz` to `attest_simulated_tpm20_test.go`, which is how it appears things work today since this test class instantiates the Linux `platformTPM`.~~

### DIFF
--- a/attest/attest.go
+++ b/attest/attest.go
@@ -19,6 +19,7 @@ import (
 	"crypto"
 	"errors"
 	"fmt"
+	"io"
 
 	"github.com/google/certificate-transparency-go/x509"
 	"github.com/google/go-tpm/tpm"
@@ -45,6 +46,7 @@ const (
 	TPMInterfaceDirect TPMInterface = iota
 	TPMInterfaceKernelManaged
 	TPMInterfaceDaemonManaged
+	TPMInterfaceCommandChannelManaged
 )
 
 // OpenConfig encapsulates settings passed to OpenTPM().
@@ -53,6 +55,10 @@ type OpenConfig struct {
 	// attempt to use. If the specified version is not available,
 	// ErrTPMNotAvailable is returned. Defaults to TPMVersionAgnostic.
 	TPMVersion TPMVersion
+	// CommandChannel can be passed if a TPM connection has already
+	// been opened, such as a simulator or a network device. Currently
+	// only supported with TPM2.0.
+	CommandChannel io.ReadWriteCloser
 }
 
 // keyEncoding indicates how an exported TPM key is represented.
@@ -85,10 +91,10 @@ const (
 )
 
 type ak interface {
-	close(*platformTPM) error
+	close(t interface{}) error
 	marshal() ([]byte, error)
-	activateCredential(tpm *platformTPM, in EncryptedCredential) ([]byte, error)
-	quote(t *platformTPM, nonce []byte, alg HashAlg) (*Quote, error)
+	activateCredential(t interface{}, in EncryptedCredential) ([]byte, error)
+	quote(t interface{}, nonce []byte, alg HashAlg) (*Quote, error)
 	attestationParameters() AttestationParameters
 }
 
@@ -395,6 +401,9 @@ func (t *probedTPM) MatchesConfig(config OpenConfig) bool {
 func OpenTPM(config *OpenConfig) (*TPM, error) {
 	if config == nil {
 		config = defaultOpenConfig
+	}
+	if config.CommandChannel != nil {
+		return openCommandChannel(config.TPMVersion, config.CommandChannel)
 	}
 	candidateTPMs, err := probeSystemTPMs()
 	if err != nil {
--- a/attest/tpm_linux.go
+++ b/attest/tpm_linux.go
@@ -18,9 +18,7 @@ package attest
 
 import (
 	"crypto"
-	"crypto/rsa"
 	"encoding/binary"
-	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -34,7 +32,6 @@ import (
 	"github.com/google/go-tspi/tspiconst"
 
 	"github.com/google/go-tpm/tpm2"
-	"github.com/google/go-tpm/tpmutil"
 	"github.com/google/go-tspi/attestation"
 )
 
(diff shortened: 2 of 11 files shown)


attest/attest.go
```diff
@@ -360,6 +365,9 @@ var (
 	// need to interact with the TPM1.2 device in ways that have not
 	// yet been implemented.
 	ErrTPM12NotImplemented = errors.New("TPM 1.2 support not yet implemented")
+	// ErrTPM20Required is returned in response to OpenTPM() when the
+	// OpenConfig CommandChannel is set and TPMVersion is TPM1.2.
+	ErrTPM20Required = errors.New("TPMVersion must be TPMVersion20 when using CommandChannel")
```
ericchiang, line 370: I don't think this needs to be part of the API. Just document that CommandChannel returns an error if TPMVersion isn't TPMVersion20.

attest/attest.go
```diff
@@ -53,6 +55,9 @@ type OpenConfig struct {
 	// attempt to use. If the specified version is not available,
 	// ErrTPMNotAvailable is returned. Defaults to TPMVersionAgnostic.
 	TPMVersion TPMVersion
+	// CommandChannel can be passed if a TPM connection has already
+	// been opened, such as a simulator or a network device
```
ericchiang, line 59: Document that only TPMVersion20 is supported when this field is provided.

attest/tpm_windows.go
```diff
@@ -95,6 +95,10 @@ func openTPM(tpm probedTPM) (*TPM, error) {
 	}}, nil
 }
 
+func openCommandChannel(rw io.ReadWriter) (*TPM, error) {
+	return nil, errors.New("CommandChannel not supported from given build parameters")
```
ericchiang, line 99: why wouldn't this be supported on windows?

attest/tpm_linux.go
```diff
@@ -116,7 +116,7 @@ func openTPM(tpm probedTPM) (*TPM, error) {
 			interf = TPMInterfaceKernelManaged
 		}
 
-		rwc, err = tpm2.OpenTPM(devPath)
+		rw, err = tpm2.OpenTPM(devPath)
```
ericchiang, line 119: This TPM is never closed. Please revert the s/io.ReadWriteCloser/io.ReadWriter/g changes. You can make Close no-op if you want

```
type noOpCloser struct {
    io.ReadWriter
}
func (n noOpCloser) Close() error {
    return
}
```

attest/tpm_linux.go
```diff
@@ -140,7 +148,11 @@ func (t *platformTPM) close() error {
 	case TPMVersion12:
 		return t.ctx.Close()
 	case TPMVersion20:
-		return t.rwc.Close()
+		rwc, ok := t.rw.(io.ReadWriteCloser)
```
ericchiang, line 151: Please revert the s/io.ReadWriteCloser/io.ReadWriter/g changes. For the CommandChallen you can make Close no-op:

```
type noOpCloser struct {
    io.ReadWriter
}
func (n noOpCloser) Close() error {
    return
}
```

attest/tpm_linux.go
```diff
@@ -126,11 +126,19 @@ func openTPM(tpm probedTPM) (*TPM, error) {
 		version: tpm.Version,
 		interf:  interf,
 		sysPath: tpm.Path,
-		rwc:     rwc,
+		rw:      rw,
 		ctx:     ctx,
 	}}, nil
 }
 
+func openCommandChannel(rw io.ReadWriter) (*TPM, error) {
```
ericchiang, line 134: Should this be in a non _linux.go file?

caleblloyd:
I refactored the changes to allow `CommandChannel` to work cross-platform.

Basically all of the platform speceifc Linux TPM 2.0 code has been pulled to a new "wrapper" called `selectorTPM`.  If `CommandChannel` is passed or Linux TPM 2.0 is used, all commands run over `io.ReadWriteCloser`.

Otherwise, a platform-specific `platformTPM` is supplied to `selectorTPM`, which calls the platform-specefic code instead.

I ran `go test` on linux and things work, but I need some help testing MacOS and Windows.  Is there a more comprehensive test suite that calls TPM 1.2 and 2.0 on Windows and Linux?

caleblloyd:
> Also how are you going to do measurement logs for emulated TPMs?

I added a check and return a not supported error for `measurementLog()` when using `CommandChannel` in the refactor

ericchiang:
I think between all of our _{os}.go files this is becoming pretty messy. I'd be really helpful to have smaller refactoring PRs before we add this feature. This PR is great but I'm having trouble reviewing with all the redundant names. Sorry we sent you on that path and happy to do some of the refactoring myself if you'd like :(

Instead of overloading platformTPM, I think we should have the following unique implementations:

* wrappedTPM - Implements using TPM 2.0 & command channel. Linux + TPM2 also uses this
* windowsTPM - windows implementation
* trousersTPM - linux implementation

For keys, we should have equivalent implementations

* windowsKey12
* windowsKey20
* trousersKey12
* wrappedKey20

I'm trying to figure out if we can avoid passing an interface{} around. It looks like the interaction between the AK interface and platformTPM makes that hard.

caleblloyd:
I agree, I tried making a common interface first but was having difficulty extracting a common interface from platformTPM that would work with keys.

If you have the cycles and want to do the refactor that is great! Otherwise I can circle back to this in a while. It's not a huge rush as I'm able to test downstream using `unsafe.Pointer` to set `platformTPM.rwc` on Linux.

PR #160, opened 2020-05-05
Refactor *platformTPM -> tpmBase interface

twitchy-jsonp:
Part 1/2 of the refactor to support injecting a wrapped TPM implementation.

This PR just moves from the existing struct pointer, where implementations where switched in using build tags, to structs implementing an interface. A future PR will:

1. Decompose `linuxTPM` into `linuxTPM12` and `wrappedTPM20`
2. Implement API surface to provided a TPM command channel instead of using the current platform
3. Move `wrappedTPM20` out of the linux build tag

### DIFF
--- a/attest/key_linux.go
+++ b/attest/key_linux.go
@@ -49,11 +49,16 @@ func (k *key12) marshal() ([]byte, error) {
 	return out.Serialize()
 }
 
-func (k *key12) close(tpm *platformTPM) error {
+func (k *key12) close(tpm tpmBase) error {
 	return nil // No state for tpm 1.2.
 }
 
-func (k *key12) activateCredential(t *platformTPM, in EncryptedCredential) ([]byte, error) {
+func (k *key12) activateCredential(tb tpmBase, in EncryptedCredential) ([]byte, error) {
+	t, ok := tb.(*linuxTPM)
+	if !ok {
+		return nil, fmt.Errorf("expected *linuxTPM, got %T", tb)
+	}
+
 	cred, err := attestation.AIKChallengeResponse(t.ctx, k.blob, in.Credential, in.Secret)
 	if err != nil {
 		return nil, fmt.Errorf("failed to activate ak: %v", err)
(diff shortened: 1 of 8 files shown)


twitchy-jsonp:
Question: There's a bunch of interface casts at the lower levels - should these be error checked, or is it sufficient to let them panic? It should be very hard for the user to get a different TPM object ...

ericchiang:
> Question: There's a bunch of interface casts at the lower levels - should these be error checked, or is it sufficient to let them panic? It should be very hard for the user to get a different TPM object ...

Maybe instead of

```go
func (k *key20) activateCredential(t tpmBase, in EncryptedCredential) ([]byte, error) {
	ekHnd, _, err := t.getPrimaryKeyHandle(commonEkEquivalentHandle)		ekHnd, _, err := t.(*linuxTPM).getPrimaryKeyHandle(commonEkEquivalentHandle)
```

Do

```go
func (k *key20) activateCredential(b tpmBase, in EncryptedCredential) ([]byte, error) {
	t, ok := b.(*linuxTPM)
	if !ok {
		return nil, fmt.Errorf("expected *linuxTPM, got %T", b)
	}
	ekHnd, _, err := t.getPrimaryKeyHandle(commonEkEquivalentHandle)		ekHnd, _, err := t.getPrimaryKeyHandle(commonEkEquivalentHandle)
```

That'd also cut down on the diff

attest/key_linux.go
```diff
@@ -49,19 +49,28 @@ func (k *key12) marshal() ([]byte, error) {
...
-func (k *key12) close(tpm *platformTPM) error {
+func (k *key12) close(tpm tpmBase) error {
 	return nil // No state for tpm 1.2.
 }
 
-func (k *key12) activateCredential(t *platformTPM, in EncryptedCredential) ([]byte, error) {
+func (k *key12) activateCredential(tb tpmBase, in EncryptedCredential) ([]byte, error) {
+	t, ok := tb.(*linuxTPM)
+	if !ok {
+		return nil, fmt.Errorf("expected *linuxTPM, got %T", t)
```
ericchiang, line 59: 't' should be 'tb'

twitchy-jsonp: Doneskies

PR #161, opened 2020-05-05
Split linuxTPM -> {trousersTPM,wrappedTPM20}, implement CommandChannel in OpenConfig

twitchy-jsonp:
Refactor PR 2/2.


attest/tpm_linux.go
```diff
@@ -118,281 +89,16 @@ func openTPM(tpm probedTPM) (*TPM, error) {
...
-			DigestAlg: alg.cryptoHash(),
-		}
-	}
-
-	return out, nil
-}
 
-func (t *linuxTPM) measurementLog() ([]byte, error) {
-	return ioutil.ReadFile("/sys/kernel/security/tpm0/binary_bios_measurements")
+	panic("unreachable")
```
ericchiang, line 103: This isn't unreachable right? It should return an error about being passed an unrecognized TPM version

attest/trousers_tpm_linux.go
```diff
@@ -0,0 +1,149 @@
+// Copyright 2020 Google Inc.
```
ericchiang, line 1: Can we keep this in tpm_linux? I don't like having even more OS specific packages.

attest/tpm_linux.go
twitchy-jsonp: It shouldnt be possible to set `version` to anything other than those two values without using unsafe (see code above, which populates `probedTPM.version`)

attest/trousers_tpm_linux.go
twitchy-jsonp: Done

attest/tpm_linux.go
twitchy-jsonp: Wait wut

twitchy-jsonp: Sec

twitchy-jsonp: Done. What the code said and what I thought it said were very different things ^^'

PR #164, opened 2020-05-07
attest: Add build flag to gate TrouSerS support

### LINKED ISSUE #163: Use build tags to allow building without tspi
Right now, a user of go-attestation must download and install libtspi-dev even if they only wish to use TPM2 (such as with GCE). This can be problematic as this library is not distributed with all Linux distros.

Go supports [build constraints](https://golang.org/pkg/go/build/#hdr-Build_Constraints) to allow for conditional compilation. Similar to how we don't require tspi on Windows builds, we could use a `notspi` build constraint to allow:
```bash
go build -notspi windows ./attest/attest-tool
```
to build without tspi.

We could also invert the semantics, having the `tspi` build flag _enable_ support for TPM 1.2 on Linux, depending on what we want to be the default.

josephlr:
Fixes #163 

The `tspi` build flag now gates support for TPM 1.2 on Linux. This
allows the attestation libraries to be built without having to first
install the TrouSerS daemon/library/headers.

A build flag of `tspi` instead of `no_tspi` was chosen as go tags tend
to be additive, but I can switch if people think having TrouSerS
support by default is a good idea. This also makes it possible for TPM 2.0
only projects to depend on go-attestation libraries.

I also updated the documentation to explain all the caveats and gotchas
when using Linux+TrouSerS+TPM 1.2

Signed-off-by: Joe Richey <joerichey@google.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,20 @@ Please note that this is not an official Google product.
 
 The go-attestation package is installable using go get: `go get github.com/google/go-attestation/attest`
 
-Linux users must install `libtspi` and its headers. This can be installed on debian-based systems using: `sudo apt-get install libtspi-dev`.
+## Linux TPM 1.2 support
+
+This package includes optional support for TPM 1.2 on Linux via the
+[TrouSerS `tcsd` daemon](http://trousers.sourceforge.net/). To build with
+support, users must:
+  - Install `libtspi` and its headers
+    - Debian-based systems: `apt install libtspi-dev`
+    - RedHat-based systems: `yum install trousers-devel`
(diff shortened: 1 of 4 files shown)


josephlr:
@twitchy-jsonp how should I go about fixing the CI. It seems to use a mixture of Google3 and GitHub files for it's configuration. Which should I change first?

twitchy-jsonp:
To prevent running the TPM1.2 tests, you just need to remove the `--testTPM12` flag from the invocation.

I expect we will want to split up the CI into different phases now that we have build invariants:

1. Existing build + test using simulator + TPM 1.2 test harness
2. Build using no_tspi and test using simulator
3. (While we are at it) check that build succeeds for `GOOS=windows`

PR #166, opened 2020-05-08
Extract properties

mjg59:
Some fixups and then the actual code to parse out the properties


mjg59:
I think it probably makes more sense to do that in a separate patch - I'll do that.

PR #167, opened 2020-05-28
Begin implementing extraction of windows info from the event log

twitchy-jsonp:
Beginnings of an extract for Windows Integrity Measurements. Theres lots more to do, but lets start here so we can get the basics right.

These arent really documented, the best I could find is [this](https://github.com/Microsoft/TSS.MSR/blob/master/PCPTool.v11/Using%20the%20Windows%208%20Platform%20Crypto%20Provider%20and%20Associated%20TPM%20Functionality.pdf), page 30. 

This PR implements:

1. Extraction of hashes which represent binary blobs which were loaded
2. Kernel security config like the NX bit, debugging mode, etc
3. Bitlocker mode

### DIFF
--- a/attest/win_events.go
+++ b/attest/win_events.go
@@ -1,0 +1,456 @@
+// Copyright 2020 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License. You may obtain a copy of
+// the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+
+package attest
+
+import (
+	"bytes"
+	"encoding/binary"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/google/go-attestation/attest/internal"
+)
+
+type windowsEvent uint32
+
+// SIPA event types
+const (
+	sipaTypeMask                    windowsEvent = 0x000f0000
+	sipaContainer                   windowsEvent = 0x00010000
+	sipaInformation                 windowsEvent = 0x00020000
+	sipaError                       windowsEvent = 0x00030000
+	sipaPreOsParameter              windowsEvent = 0x00040000
+	sipaOSParameter                 windowsEvent = 0x00050000
+	sipaAuthority                   windowsEvent = 0x00060000
+	sipaLoadedModule                windowsEvent = 0x00070000
+	sipaTrustPoint                  windowsEvent = 0x00080000
+	sipaELAM                        windowsEvent = 0x00090000
+	sipaVBS                         windowsEvent = 0x000a0000
+	trustBoundary                   windowsEvent = 0x40010001
+	elamAggregation                 windowsEvent = 0x40010002
+	loadedModuleAggregation         windowsEvent = 0x40010003
+	trustpointAggregation           windowsEvent = 0xC0010004
+	ksrAggregation                  windowsEvent = 0x40010005
+	ksrSignedMeasurementAggregation windowsEvent = 0x40010006
+	information                     windowsEvent = 0x00020001
+	bootCounter                     windowsEvent = 0x00020002
+	transferControl                 windowsEvent = 0x00020003
+	applicationReturn               windowsEvent = 0x00020004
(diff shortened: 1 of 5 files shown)


twitchy-jsonp:
Heh, CI doesnt have Go 1.13. standby

attest/win_events.go
```diff
@@ -0,0 +1,444 @@
...
+		}
+
+		et, err := internal.UntrustedParseEventType(uint32(e.Type))
+		if err != nil {
+			return nil, fmt.Errorf("unrecognised event type: %v", err)
+		}
+
+		digestVerify := e.digestEquals(e.Data)
+		switch et {
+		case internal.EventTag:
```
twitchy-jsonp, line 248: Dont forget to fail if this is not the event type

attest/win_events.go
```diff
@@ -0,0 +1,444 @@
...
+	if err := binary.Read(r, binary.LittleEndian, &header); err != nil {
+		return sipaEvent{}, err
+	}
+
+	// fmt.Printf("ID: %x (%v)\n", header.Type, header.Type&trustBoundary == trustBoundary)
+
+	switch header.Type {
+	// Parse byte sequences.
+	case authenticodeHash:
+		data := make([]byte, header.Size)
```
ericchiang, line 341: same note here about allocating unbounded amount of memory

attest/win_events.go
```diff
@@ -0,0 +1,462 @@
...
+// unless the integrity of platform & bootloader events has already been
+// established.
+func ParseWinEvents(events []Event) (*WinEvents, error) {
+	var (
+		out           = WinEvents{LoadedModules: map[string]WinModuleLoad{}}
+		seenSeparator bool
+	)
+
+	for _, e := range events {
+		//fmt.Println(e)
```
ericchiang, line 189: nit: remove

attest/win_events.go
```diff
@@ -0,0 +1,462 @@
...
+
+	// All windows information should be sub events in an enclosing SIPA
+	// container event.
+	if (windowsEvent(evt.ID) & sipaTypeMask) != sipaContainer {
+		return fmt.Errorf("expected container event, got %v", windowsEvent(evt.ID))
+	}
+
+eventLoop:
+	for {
+		if err := w.readSIPAEvent(r); err != nil {
```
ericchiang, line 450: why isn't this

```
for r.Len() > 0 {
    if err := w.readSIPAEvent(r); err != nil {
        if errors.Is(err, unknownSIPAEvent) {
            // Unknown SIPA events are okay as all TCG events are verifiable.
            continue eventLoop
        }
        return err
    }
}
return nil
```

instead of doing error handling for EOF in the loop?

attest/win_events.go
```diff
@@ -0,0 +1,462 @@
...
+	return nil
+}
+
+func (w *WinEvents) readSIPAEvent(r io.Reader) error {
+	var header microsoftEventHeader
+	if err := binary.Read(r, binary.LittleEndian, &header); err != nil {
+		return err
+	}
+
+	// fmt.Printf("ID: %x (%v)\n", header.Type, header.Type&trustBoundary == trustBoundary)
```
ericchiang, line 409: nit: remove

attest/win_events.go
```diff
@@ -0,0 +1,462 @@
...
+			return fmt.Errorf("unknown event in LMA aggregation: %v", header.Type)
+		}
+	}
+
+	l := WinModuleLoad{
+		AuthenticodeHash: codeHash,
+		ImageBase:        []uint64{imgBase},
+	}
+	hashHex := hex.EncodeToString(l.AuthenticodeHash)
+	l.ImageBase = append(l.ImageBase, w.LoadedModules[hashHex].ImageBase...)
```
ericchiang, line 398: need to ensure hashHex is actually in LoadedModules?
```
m, ok := w.LoadedModules[hashHex]
if !ok {
    // ...
}
l.ImageBase = append(l.ImageBase, m.ImageBase...)
```

attest/win_events_test.go
```diff
@@ -0,0 +1,327 @@
...
+	if err != nil {
+		t.Fatalf("validating event log: %v", err)
+	}
+
+	winState, err := ParseWinEvents(events)
+	if err != nil {
+		t.Fatalf("ExtractSecurebootState() failed: %v", err)
+	}
+
+	// var s strings.Builder
```
ericchiang, line 314: Remove this?

attest/win_events.go
twitchy-jsonp: Nack, the `nil` default value of `ImageBase` passed to `append()` should be safe, as `nil...` variadically expands to 0 arguments.

ericchiang:
(please squash before mergin)

PR #168, opened 2020-05-29
Handle platform certificates that only provide a single property

mjg59:
The spec states that the PlatformProperties field of the
PlatformConfiguration attribute should be a sequence of key value pairs.
However, it seems that if there's only a single property present, it's
sometimes being stored as a bare key value pair rather than a sequence
with a single entry. Work around that.

### DIFF
--- a/attributecert/attributecert.go
+++ b/attributecert/attributecert.go
@@ -448,6 +448,13 @@ type PlatformConfigurationV2 struct {
 	PlatformPropertiesURI   URIReference            `asn1:"optional,tag:3"`
 }
 
+type PlatformConfigurationV2_workaround struct {
+	ComponentIdentifiers    []ComponentIdentifierV2 `asn1:"optional,tag:0"`
+	ComponentIdentifiersURI URIReference            `asn1:"optional,tag:1"`
+	PlatformProperty        Property                `asn1:"optional,tag:2"`
+	PlatformPropertiesURI   URIReference            `asn1:"optional,tag:3"`
+}
+
 type ComponentIdentifierV1 struct {
 	ComponentClass          []byte `asn1:"optional"`
 	ComponentManufacturer   string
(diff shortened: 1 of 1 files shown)


attributecert/attributecert.go
```diff
@@ -448,6 +448,13 @@ type PlatformConfigurationV2 struct {
 	PlatformPropertiesURI   URIReference            `asn1:"optional,tag:3"`
 }
 
+type PlatformConfigurationV2_workaround struct {
```
twitchy-jsonp, line 451: Use CamelCase: `PlatformConfigurationV2Workaround` or  `PlatformConfiguration2Workaround`

attributecert/attributecert.go
```diff
@@ -580,7 +587,14 @@ func parseAttributeCertificate(in *attributeCertificate) (*AttributeCertificate,
 		case attribute.ID.Equal(oidTcgPlatformConfigurationV2):
 			var platformConfiguration PlatformConfigurationV2
 			if _, err := asn1.Unmarshal(attribute.RawValues[0].FullBytes, &platformConfiguration); err != nil {
-				return nil, err
+				var workaround PlatformConfigurationV2_workaround
+				if _, err := asn1.Unmarshal(attribute.RawValues[0].FullBytes, &workaround); err != nil {
+					return nil, err
```
twitchy-jsonp, line 592: Is the error message specific enough that its clear its coming from the workaround unmarshal?

If not, can this be `return nil, fmt.Errorf("parsing workaround: %v", err)`

attributecert/attributecert.go
mjg59: Done

attributecert/attributecert.go
mjg59: We don't return a specific message for any parsing failures (which is arguably bad but is also how the upstream x509 code behaves). I think fixing that up is probably best done in a different patchset.

PR #173, opened 2020-06-18
windows events: Extract the ELAM driver names and their configuration

twitchy-jsonp:


attest/win_events.go
```diff
@@ -394,13 +409,93 @@ eventLoop:
...
+		measured   []byte
+		policy     []byte
+		config     []byte
+	)
+
+eventLoop:
+	for {
+		var header microsoftEventHeader
+		if err := binary.Read(r, binary.LittleEndian, &header); err != nil {
+			if errors.Is(err, io.EOF) {
```
ericchiang, line 435: I think we discussed this in the last PR, but why isn't this

```
r = io.LimitedReader{R: rdr, N: int64(header.Size)}

for r.N > 0 {
    // ...
}
```

io.EOF can mean that there's trialing data that we probably shouldn't tolerate.

attest/win_events.go
```diff
@@ -394,13 +409,93 @@ eventLoop:
...
+		r          = io.LimitReader(rdr, int64(header.Size))
+		driverName string
+		measured   []byte
+		policy     []byte
+		config     []byte
+	)
+
+eventLoop:
+	for {
+		var header microsoftEventHeader
```
ericchiang, line 433: Rename this to `h` or something similar, it's confusing that this is uses the same name as the function argument.

attest/win_events.go
```diff
@@ -394,13 +409,93 @@ eventLoop:
...
+		if err := binary.Read(r, binary.LittleEndian, &header); err != nil {
+			if errors.Is(err, io.EOF) {
+				break eventLoop
+			}
+			return fmt.Errorf("parsing ELAM aggregation sub-event: %v", err)
+		}
+		if int(header.Size) > rdr.Len() {
```
ericchiang, line 440: This should be r.N instead of rdr.Len(), right? Since reading more than the limited reader will produce an error below.

attest/win_events.go
```diff
@@ -394,13 +409,93 @@ eventLoop:
...
+			config = make([]byte, header.Size)
+			if err := binary.Read(r, binary.LittleEndian, &config); err != nil {
+				return fmt.Errorf("reading ELAM config value: %v", err)
+			}
+		default:
+			return fmt.Errorf("unknown event in LMA aggregation: %v", header.Type)
+		}
+	}
+
+	w.ELAM[driverName] = WinELAM{
```
ericchiang, line 482: add a check here to make sure a driverName was provided.

```
if driverName == "" {
    return fmt.Errorf("no name for driver") // or whatever
}
```

attest/win_events.go
```diff
@@ -175,13 +180,23 @@ type WinModuleLoad struct {
 	ImageBase []uint64
 }
 
+// WinELAM describes the configuration of an Early Launch AntiMalware driver.
+type WinELAM struct {
+	Measured []byte
```
ericchiang, line 185: Add a short comment indicating what these fields are?

attest/win_events.go
twitchy-jsonp: Done, I also fixed up this and related comments for a similar function `readLoadedModuleAggregation()`.

PR #174, opened 2020-07-23
Create oid package to centralize TCG relevant OIDs

brandonweeks:
This allows for sharing of OID definitions between the existing `attributecert` package and the future `ekcert` package. Identifiers not in the `id-ce`/`tcg` hierarchies or those that are obsolete remain in the `attributecert` package.

### DIFF
--- a/oid/oid.go
+++ b/oid/oid.go
@@ -1,0 +1,48 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


oid/oid.go
```diff
@@ -0,0 +1,34 @@
+// Package oid contains X.509 and TCG ASN.1 object identifiers.
```
ericchiang, line 1: Need a Google copyright header?

brandonweeks: Done.

PR #179, opened 2020-09-24
Disable linux TPM12 support if build without cgo

### LINKED ISSUE #163: Use build tags to allow building without tspi

twitchy-jsonp:
Fixes #163
Closes #164 

This PR absorbs #164 (thanks @josephlr) but switches the build tags to use builtin ones (`cgo`) instead of adding `tspi` or `no_tspi`. If the user builds with cgo functioning, then linux TPM12 is enabled. Otherwise, only TPM 2.0 is supported on linux (but both versions supported on Windows).

### DIFF
--- a/attest/key_linux.go
+++ b/attest/key_linux.go
@@ -12,7 +12,7 @@
 // License for the specific language governing permissions and limitations under
 // the License.
 
-// +build linux,!gofuzz
+// +build linux,!gofuzz,cgo
 
 package attest
 
--- a/attest/tpm12_linux.go
+++ b/attest/tpm12_linux.go
@@ -1,0 +1,164 @@
+// Copyright 2020 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License. You may obtain a copy of
+// the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


ericchiang (approved):
lgtm

josephlr (approved):
Looks good, flagging on cgo makes sense here.

lihanshang:
Hey Guys
I am running a test and I am seeing 
```
# github.com/google/go-tspi/tspi
/go/pkg/mod/github.com/google/go-tspi@v0.2.1-0.20190423175329-115dea689aad/tspi/context.go:17:27: fatal error: trousers/tss.h: No such file or directory
 // #include <trousers/tss.h>
                           ^
compilation terminated.
```
I am assuming this is the related issue of not installing the tspi-dev since I am running it on centos. May I know if I dont want TPM12 support what I should run to build? 

Thanks!

PR #182, opened 2020-11-17
Add missing defaults

mjg59:
Some fields have default values, which also means they should be treated
as optional. Fix up the tagging to handle that.

### DIFF
--- a/attributecert/attributecert.go
+++ b/attributecert/attributecert.go
@@ -320,7 +320,7 @@ type TCGCredentialType struct {
 type FipsLevel struct {
 	Version string
 	Level   asn1.Enumerated
-	Plus    bool
+	Plus    bool `asn1:"optional,default=false"`
 }
 
 type CommonCriteriaMeasures struct {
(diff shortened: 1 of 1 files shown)


attributecert/attributecert.go
```diff
@@ -336,11 +336,11 @@ type CommonCriteriaMeasures struct {
 }
 
 type TBBSecurityAssertions struct {
-	Version          int
+	Version          int                    `asn1:"optional,default=0"`
```
ericchiang, line 339: I think this should have to be ':' not '='

Also ints already default to 0. If you mark optional, is that enoguh?

attributecert/attributecert.go
```diff
@@ -320,7 +320,7 @@ type TCGCredentialType struct {
 type FipsLevel struct {
 	Version string
 	Level   asn1.Enumerated
-	Plus    bool
+	Plus    bool `asn1:"optional,default=false"`
```
ericchiang, line 323: Similar to the comment about Version, bool should already default to false. Is marking as optional enough?

mjg59:
These appear to be fine just tagged as optional

attributecert/attributecert.go
ericchiang: I'm still seeing "default=0" in my diff. Is this intentional? :)

twitchy-jsonp (approved):

PR #185, opened 2020-11-19
Fix Go Vet string conversion warnings

brandonweeks:
https://golang.org/doc/go1.15#vet


attest/tpm.go
```diff
@@ -121,7 +121,7 @@ func readTPM2VendorAttributes(tpm io.ReadWriter) (tpm20Info, error) {
 			return tpm20Info{}, fmt.Errorf("got capability of type %T, want tpm2.TaggedProperty", caps[0])
 		}
 		// Reconstruct the 4 ASCII octets from the uint32 value.
-		vendorInfo += string(subset.Value&0xFF000000) + string(subset.Value&0xFF0000) + string(subset.Value&0xFF00) + string(subset.Value&0xFF)
```
twitchy-jsonp, line 124: Don't we want to actually intepret this as a string thou?

ie: `utf8.EncodeRune()` or `string(rune())`

twitchy-jsonp (approved):

PR #189, opened 2020-11-25
Export InvalidPCRs field in ReplayError

aditya4k:
## Proposed Changes 
In order to retrieve the Invalid PCRs which couldn't be replayed against the Event log, we need this field to be exported as this gives the exact and true information. Replay error events will give all the events, but doesn't give the exact PCR index which doesn't get replayed. 

Following is the test to extend PCR 7 and verify the PCRs 7,8,9 against the Event log. Output:
```
event log failed to verify: the following registers failed to replay: [7]
ReplayError Events:=[107]
Replay Error Events PCR indexes=[0 7 2 3 6 9 8 1 4 5]
```

### DIFF
--- a/attest/eventlog.go
+++ b/attest/eventlog.go
@@ -38,11 +38,11 @@ import (
 // a particular PCR.
 type ReplayError struct {
 	Events      []Event
-	invalidPCRs []int
+	InvalidPCRs []int
 }
 
 func (e ReplayError) affected(pcr int) bool {
-	for _, p := range e.invalidPCRs {
+	for _, p := range e.InvalidPCRs {
 		if p == pcr {
 			return true
 		}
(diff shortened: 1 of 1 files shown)


aditya4k:
@googlebot I signed it!

aditya4k:
@ericchiang  could you please check if this change sounds reasonable?

ericchiang:
Seems reasonable to me. If you're going to export this, can you include a comment above it? E.g.

```
// InvalidPCRs reports the set of PCRs where the event log replay failed.
```

aditya4k:
@ericchiang updated the Field with the Comment. Please have a look. Thanks

ericchiang (approved):
lgtm, reassigning to @twitchy-jsonp since he's been working the most with this lately

aditya4k:
@twitchy-jsonp  could you please have a look at it please. Thanks

twitchy-jsonp:
I've kicked the CI. Will merge once it reports green.

PR #193, opened 2020-12-01
Deepikarajani24 patch 1

deepikarajani24:
this PR removed the dependency on tbs.dll when TPM support is not required for windows which is the case for any independent verifier application which doesn't need to have tbs.dll available to do the verification of an attest machine

### DIFF
--- a/attest/tpm_windows.go
+++ b/attest/tpm_windows.go
@@ -42,6 +42,10 @@ type windowsTPM struct {
 func (*windowsTPM) isTPMBase() {}
 
 func probeSystemTPMs() ([]probedTPM, error) {
+	// Initialize Tbs.dll here so that it's linked only when TPM support is required.
+	tbs = windows.MustLoadDLL("Tbs.dll")
+	tbsGetDeviceInfo = tbs.MustFindProc("Tbsi_GetDeviceInfo")
+	
 	// Windows systems appear to only support a single abstracted TPM.
 	// If we fail to initialize the Platform Crypto Provider, we assume
 	// a TPM is not present.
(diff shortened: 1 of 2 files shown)


attest/tpm_windows.go
```diff
@@ -42,6 +42,10 @@ type windowsTPM struct {
 func (*windowsTPM) isTPMBase() {}
 
 func probeSystemTPMs() ([]probedTPM, error) {
+	// Initialize Tbs.dll here so that it's linked only when TPM support is required.
+	tbs = windows.MustLoadDLL("Tbs.dll")
```
twitchy-jsonp, line 46: Can you surround this in a check to only load the DLL if it hasnt already been loaded? ie: `tbs == nil`

twitchy-jsonp (changes requested):
Looks good with one comment. Should be just about g2g once we get the CLA bot passing!

attest/tpm_windows.go
deepikarajani24: sure, thanks!

deepikarajani24:
@googlebot I signed it!

attest/tpm_windows.go
```diff
@@ -42,6 +42,12 @@ type windowsTPM struct {
 func (*windowsTPM) isTPMBase() {}
 
 func probeSystemTPMs() ([]probedTPM, error) {
+	// Initialize Tbs.dll here so that it's linked only when TPM support is required.
+	if tbs == nil {
+		tbs = windows.MustLoadDLL("Tbs.dll")
+	}
+	tbsGetDeviceInfo = tbs.MustFindProc("Tbsi_GetDeviceInfo")
```
twitchy-jsonp, line 49: Move this into the if statement as well (its duplicated work for the procedure to be loaded twice - these globals should be treated as singletons)

twitchy-jsonp (commented):
One last comment then I can merge :)

twitchy-jsonp (approved):

PR #199, opened 2021-01-08
attestPCRs(): make sure that the return values are consistent

pszal:


pszal:
@googlebot I signed it!

ericchiang (approved):

PR #201, opened 2021-02-16
Support for application signing keys

pszal:
This PR allows for creating (and verifying) non-restricted TPM-backed keys which are certified by an attestation key. These keys can be used by applications for signing outside-TPM objects.


pszal:
It depends on https://github.com/google/go-tpm-tools/pull/99, but except that, it is ready for review and feedback.

ericchiang:
Thanks! Would you mind including a description of the next exported APIs?

```
type SK AK

func (s *SK) Private(pub crypto.PublicKey) (crypto.PrivateKey, error)
```

Or whatever... we probably want to discuss that first.

Also tpmtools is a pretty large import. If we can we probably want to avoid that outside of tests.

cc @twitchy-jsonp

attest/activation.go
```diff
@@ -31,6 +31,17 @@ const (
...
 
+// secureCurves represents a set of secure elliptic curves. For now,
+// the selection is based on the key size only.
+var secureCurves = map[tpm2.EllipticCurve]bool{
+	tpm2.CurveNISTP256: true,
+	tpm2.CurveNISTP384: true,
+	tpm2.CurveNISTP521: true,
+	tpm2.CurveBNP256:   true,
+	tpm2.CurveBNP638:   true,
+	tpm2.CurveSM2P256:  true,
```
brandonweeks, line 42: I doubt we think this is secure.

brandonweeks (changes requested):

attest/activation.go
pszal: Dropped.

attest/activation.go
```diff
@@ -88,32 +98,68 @@ func (p *ActivationParameters) checkTPM12AKParameters() error {
 	return nil
 }
 
+// VerifyOpts specifies options passed to (*AttestationParameters).Verify()
+type VerifyOpts struct {
+	// SelfAttested set to true ensures that the attestation is self-signed,
+	// set to false ensures that the attestation is not self-signed.
+	SelfAttested bool
```
ericchiang, line 105: This should be "InsecureSelfAttested" similar to "InsecureSkipVerify" https://golang.org/pkg/crypto/tls/#Config

Is there a reason we wanted to support this?

attest/activation.go
```diff
@@ -88,32 +98,68 @@ func (p *ActivationParameters) checkTPM12AKParameters() error {
...
+	// Restricted set to true ensures that the verified key was created by TPM
```
ericchiang, line 106: If we're going to support this, please add a note that the SK won't work as a crypto.Signer if it's restricted.

Also do we actually need this for something? Maybe just drop it entirely?

attest/activation.go
```diff
@@ -88,32 +98,68 @@ func (p *ActivationParameters) checkTPM12AKParameters() error {
...
+}
+
+// Verify verifies the TPM2-produced attestation parameters checking whether:
+// - the attestation is self-signed (specified by opts)
+// - the key length is secure
+// - the attestation parameters matched the attested key
+// - the key was TPM-generated and resides within TPM
+// - the key can or cannot sing outside TPM-objects (specified via opts)
+// - the signature is correct
+func (p *AttestationParameters) Verify(opts VerifyOpts) error {
```
ericchiang, line 127: The way this is laid out, it's really hard to see what's being changed for "Generate()" or what data is verified by the signature in "Verify()". Also if any of this is wrong, that's an auth bypass :)

I also don't see the reason we want to check the signature for Generate(), that seems redundant?

I would have expected logic like

- Generate()
  - check key attributes
  - generate encrypted challenge
- Verify()
  - check signature
  - check key attributes

Where the "check key attributes" part is its own function separately called by Generate() and Verify(). That makes it clear what is common, but ignore other components like "SelfAttested"?

attest/activation.go
```diff
@@ -160,28 +215,25 @@ func (p *ActivationParameters) checkTPM20AKParameters() error {
...
-	signHash, err := pub.RSAParameters.Sign.Hash.Hash()
+	// TODO: Support ECC certifying keys
+	pk := rsa.PublicKey{E: int(vrfy.RSAParameters.Exponent()), N: vrfy.RSAParameters.Modulus()}
+	signHash, err := vrfy.RSAParameters.Sign.Hash.Hash()
 	if err != nil {
 		return err
 	}
 	hsh := signHash.New()
-	hsh.Write(p.AK.CreateAttestation)
-	verifyHash, err := pub.RSAParameters.Sign.Hash.Hash()
```
ericchiang, line 170: Why did this go away? It's not clear to me what's being verified or why.

attest/attest.go
```diff
@@ -201,6 +203,10 @@ type AttestationParameters struct {
 	// the TPM Part 2 Structures specification, available at
 	// https://trustedcomputinggroup.org/wp-content/uploads/TPM-Main-Part-2-TPM-Structures_v1.2_rev116_01032011.pdf
 
+	// CertifyingKey is a public key used for the certification of Public. Only
+	// TPM 2.0 is supported and the key is encoded as a TPMT_PUBLIC structure.
+	CertifyingKey []byte
```
ericchiang, line 208: AttestationParameters are generated by the client and attacker controlled. I think we should just drop this field as isn't never verified and basically should never be used. Make this an argument of VerifyOpts?

attest/activation.go
ericchiang: Realize this may be a historical issue rather than an issue with this PR. @twitchy-jsonp if he has any thoughts on how we might make this a little cleaner and easier to verify it's correct.

attest/activation.go
```diff
@@ -88,32 +98,68 @@ func (p *ActivationParameters) checkTPM12AKParameters() error {
 	return nil
 }
 
+// VerifyOpts specifies options passed to (*AttestationParameters).Verify()
+type VerifyOpts struct {
```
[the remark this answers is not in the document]
pszal: Yes, AttestationParmeters has the CertifyingKey field.

attest/activation.go
pszal: AK's AttestationParameters are self-signed, so I keep it for this case. I'm fine with marking it insecure.

attest/activation.go
pszal: For AK's AttestationParameters this check is required. I'll add a note.

attest/activation.go
pszal: I'm not sure are we on the same page. The Generate()/Verify() logic is for the credential activation protocol, involving ActivationParameters (with a challenge, AttestationParameters,...). AttestationParameters' verification is only part of that (as mentioned, AttestationParameter's Verify for this case must set both options SelfSigned and Restricted to true)

Key certification is non-interactive: client/prover generates `ap=sk.AttestationParameters()` and server/verifier just calls `ap.Verify(...)` (SelfSigner=Restricted=false) and makes sure `ap.CertifyingKey` is trusted.

attest/activation.go
ericchiang: I guess my point is this shouldn't be exposed to the user. We don't require the user to call Verify() for AKs today, and I don't think it's reasonable to start requiring it now.

attest/activation.go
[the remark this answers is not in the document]
ericchiang: Saw that after leaving this comment :) As stated, I think that's the wrong place to put it. It's very similar to JWTs with a public key in the header. You can _never_ trust it or use it correctly, so it's probably better not to expose to the user.

attest/activation.go
ericchiang: Same as my other comment, this probably shouldn't be exposed to the user then. We can't require a new call since it's not required today.

attest/activation.go
pszal: Before, it was hardcoded to verify self-signed attestation. Now it is the same logic by the signature is verified against CertyfingKey.
`verifyHash` was redundant, wasn't it?

attest/attest.go
pszal: Certifying key is a parameter of the key certification process, so it felt natural to include it within AttestationParameters. Yes, the verifier needs to make sure that CertifyingKey is trusted.

attest/activation.go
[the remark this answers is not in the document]
pszal: Makes sense! I'll make it a parameter then.

attest/activation.go
pszal: Where a new call would be required?

But for hiding options from callers, what about renaming it to `verify(...)` and then expose `VerifyAppKey(pub)` that would call `verify(...)` with Restricted=false?

attest/certification.go
```diff
@@ -0,0 +1,169 @@
...
+	// CreateSignature represents a signature of the CreateAttestation structure.
+	// It is encoded as a TPMT_SIGNATURE structure.
+	CreateSignature []byte
+}
+
+// VerifyOpts specifies options for the key certification's verification.
+type VerifyOpts struct {
+	// Public is the public key used to verify key ceritification.
+	Public crypto.PublicKey
+	// Has is the hash function used for signature verification.
```
ericchiang, line 59: s/Has/Hash/g

Also where does the user get this hash? From the verifying key? Can you please expand the comment?

attest/application_key_test.go
```diff
@@ -0,0 +1,162 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+
+// +build !localtest !tpm12
+// +build cgo
+
+// NOTE: simulator requires cgo, hence the build tag.
```
ericchiang, line 18: nit: comments right above the "package" declaration are usually reserved for godoc. Add a newline.

attest/application_key.go
```diff
@@ -0,0 +1,103 @@
...
+	close(tpmBase) error
+	marshal() ([]byte, error)
+	certificationParameters() CertificationParameters
+	sign(tpmBase, []byte) ([]byte, error)
+	decrypt(tpmBase, []byte) ([]byte, error)
+}
+
+// ApplicationKey represents a key which can be used for signing and decrypting
+// outside-TPM objects.
+type ApplicationKey struct {
```
ericchiang, line 35: nit: not a huge deal, but I noticed that ApplicationKey seems like a long name. Same with NewAppKey, LoadAppKey, etc. What if we just named this "Key"? So it'd be

- NewKey()
- LoadKey()
- KeyConfig
- etc.

Does that seem easier? Also fine with just sticking with ApplicationKey.

attest/application_key_test.go
```diff
@@ -0,0 +1,162 @@
...
+	if !ok {
+		t.Fatalf("want crypto.Signer, got %T", priv2)
+	}
+	pk2, err := x509.MarshalPKIXPublicKey(signer2.Public())
+	if err != nil {
+		t.Fatalf("cannot marshal public key: %v", err)
+	}
+
+	if !bytes.Equal(pk1, pk2) {
+		t.Error("public keys do noy match")
```
ericchiang, line 107: s/noy/not/g

attest/tpm.go
```diff
@@ -335,6 +354,19 @@ func (t *TPM) NewAK(opts *AKConfig) (*AK, error) {
 	return t.tpm.newAK(opts)
 }
 
+// NewAppKey creates an application key certified by the attestation key ak.
```
ericchiang, line 357: nit: drop "ak" at the end or put it in parentheses or capitalize 

```
NewAppKey creates an application key certified by the attestation key (AK).
```

ericchiang (approved):

ericchiang:
lgtm, just nits. Please squash your commits before merging :)

attest/application_key.go
pszal: `Key` sounds much better, renamed ;-).
All other comments are addressed too.

brandonweeks (approved):

mjg59:
Would it be possible to cut a release containing this PR?

brandonweeks:
I've been meaning to fully sync our internal patches with GitHub. I'll do that now and cut a release.

brandonweeks:
https://github.com/google/go-attestation/releases/tag/v0.2.3

PR #207, opened 2021-03-31
Minor fixes and additions

pszal:
- replace ReadPublic() by DecodePublic() when creating and loading keys: the current implementation calls ReadPublic() even if public data is already accessible
- drop handle() from the ak interface: it is unnecessary
- add PrivPubBlobs() to attest.Key: to allow `attest`-agnostic  key marshaling

### DIFF
--- a/attest/application_key.go
+++ b/attest/application_key.go
@@ -103,7 +103,7 @@ func (k *Key) CertificationParameters() CertificationParameters {
 	return k.key.certificationParameters()
 }
 
-// PrivPubBlobs returns private and public blobs to be used by tpm2.Load().
-func (k *Key) PrivPubBlobs() ([]byte, []byte, error) {
+// PubPrivBlobs returns private and public blobs to be used by tpm2.Load().
+func (k *Key) PubPrivBlobs() ([]byte, []byte, error) {
 	return k.key.blobs()
 }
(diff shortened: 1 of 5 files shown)


attest/application_key.go
```diff
@@ -101,3 +102,8 @@ func (k *Key) Marshal() ([]byte, error) {
 func (k *Key) CertificationParameters() CertificationParameters {
 	return k.key.certificationParameters()
 }
+
+// PubPrivBlobs returns private and public blobs to be used by tpm2.Load().
+func (k *Key) PubPrivBlobs() ([]byte, []byte, error) {
```
ericchiang, line 107: Want to just call this "Blobs()"? Or "Export()"?

ericchiang (approved):

attest/application_key.go
```diff
@@ -101,3 +102,8 @@ func (k *Key) Marshal() ([]byte, error) {
 func (k *Key) CertificationParameters() CertificationParameters {
 	return k.key.certificationParameters()
 }
+
+// Blobs returns private and public blobs to be used by tpm2.Load().
+func (k *Key) Blobs() ([]byte, []byte, error) {
```
ericchiang, line 107: Blah, sorry I didn't catch this. But please name the return values. 

func (k *Key) Blobs() (pub, priv []byte, err error)

attest/application_key.go
pszal: Done

attest/application_key.go
pszal: Done

PR #210, opened 2021-04-09
Add AK.Certify() and use CertifyEx() for certification

pszal:
- replace CertifyCreation() by CertifyEx() to handle certification of objects for which we cannot extract CreationData
- add AK.Certify(handle) allowing to certify externally-created keys
- ~~small refactoring of CertificationParameters~~


attest/certification.go
```diff
@@ -41,15 +43,12 @@ type CertificationParameters struct {
...
 	Public []byte
-	// CreateData represents the properties of a TPM 2.0 key. It is encoded
-	// as a TPMS_CREATION_DATA structure.
-	CreateData []byte
-	// CreateAttestation represents an assertion as to the details of the key.
+	// Attestation represents an assertion as to the details of the key.
 	// It is encoded as a TPMS_ATTEST structure.
-	CreateAttestation []byte
-	// CreateSignature represents a signature of the CreateAttestation structure.
+	Attestation []byte
```
ericchiang, line 48: We generally avoid all breaking changes to APIs like this. This can't change unless there's a _very_ good reason to.

attest/attest.go
```diff
@@ -145,6 +146,14 @@ func (k *AK) AttestationParameters() AttestationParameters {
 	return k.ak.attestationParameters()
 }
 
+// Certify uses the attestation key to certify the key with `handle`. It returns
+// certification parameters which allow to verify the properties of the attested
+// key. Depending on the actual instantiation it can accept different handle
+// types (e.g., tpmutil.Handle or uintptr).
+func (k *AK) Certify(tb tpmBase, handle interface{}) (*CertificationParameters, error) {
```
ericchiang, line 153: This probably shouldn't be an interface? Where does the handle come from? Is it from the same process? Can we expose a "type Handle struct" with internal fields?

attest/certification.go
pszal: Reverted. (Although I thought it is ok to clean it now when it is only in a draft release.)

attest/attest.go
pszal: handle will come from this package (tpmutil.Handle) or from Windows' certtostore (uintrptr).
You mean something like this?
<pre>
type Handle struct {
  Native tpmutil.Handle
  PCP uintrptr
}
</pre>

attest/attest.go
```diff
@@ -97,12 +98,19 @@ const (
 	keyEncodingParameterized
 )
 
+// Handle represents different types of a key handle.
+type Handle struct {
+	Native *tpmutil.Handle
```
ericchiang, line 103: I think this should either be an opaque handle with private fields, or just use an empty interface{}.

E.g. this should either be

type Handle struct {
    // contains private fields
}

Does this need to interface with other code or can it just be local to go-attestation? Basically does anyone else ever need to import a handle?

How do you expect to actually call Certify(), for example?

pszal: It is needed for the case when a key created outside go-attestation has to be certified. For instance, a win pcp `key` is created with `certtostore` and we would like to get it certified by our `attestClient` (with `ak *AK`). Then I expect to call something like `attestClient.Certify(Handle{PCP:key.Handle()})` that would call `ak.Certify(..)`. 

I'm equally ok with the current `Handle` or
```
type Handle struct {
   Value interface{}
}
```

ericchiang: Thanks for the info. The Handle suggestion was for type safety. If that's the case, we should either have constructors for Handle:

```
type Handle struct {
    // private field
}
func TLS12Handle(h uintptr) *Handle {
    // ...
}
func TLS20Handle(h tpmutil.Handle) *Handle {
    // ...
}
```

Or just have Certify take the handle then type switch (which is what you did originally):

```
// For  a TPM 1.2 handle pass a uintptr, for a TPM 2.0 handle pass a tpmutil.Handle.
func (...) Certify(handle interface{}) ... {

}
```

Given that I think I'd prefer the latter? Just do the type switch.

pszal: Done. (I do type checking in the underlying ak.Certify()).

ericchiang (approved):
lgtm. please squash your commits before merging

PR #212, opened 2021-04-14
Parse TCG_PCClientPCREvent structures with an eventSize of 0

brandonweeks:
Some TPM 1.2 devices emit `EV_IPL` events with an `eventSize` of 0 and an empty `event` field. The specification would appear to suggest that only the digest itself is relevant for this event type and the information stored in the `event` field is only informational.

From [TCG PC Client Specific Implementation Specification for Conventional BIOS](https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClientImplementation_1-21_1_00.pdf):
```
Used in PCR[4] and may be used in OS PCRs (e.g.,
8-15).
The digest field contains the SHA-1 hash of the IPL
Code. The event field SHOULD NOT contain the
actual IPL Code but MAY contain informative
information about the IPL Code. Note: The digest
may not cover the entire area hosting the IPL Image,
but only the portion that contains the IPL Code. For
example, if the IPL Image is a disk drive MBR, this
MUST NOT include the portion of the MBR that
contains the disk geometry.
For MBR code, the event data SHOULD be “MBR”.
For floppy disks, the event data SHOULD be
“FLOPPY DISK”.
For BCV devices, the event data SHOULD be “BCV
IPL”.
For BEV devices, the event data SHOULD be “BEV
IPL”.
For a PARTIES partition, the event data SHOULD be
“PARTIES IPL”.
For El Torito complaint devices, the event data
SHOULD be “EL TORITO IPL”.
For other devices, the event data SHOULD be “IPL”.
See Section 3.3.3.5.```

### DIFF
--- a/attest/eventlog.go
+++ b/attest/eventlog.go
@@ -657,9 +657,6 @@ func parseRawEvent(r *bytes.Buffer, specID *specIDEvent) (event rawEvent, err er
 	if err = binary.Read(r, binary.LittleEndian, &h); err != nil {
 		return event, fmt.Errorf("header deserialization error: %w", err)
 	}
-	if h.EventSize == 0 {
-		return event, errors.New("event data size is 0")
-	}
 	if h.EventSize > uint32(r.Len()) {
 		return event, &eventSizeErr{h.EventSize, r.Len()}
 	}
--- a/attest/eventlog_test.go
+++ b/attest/eventlog_test.go
@@ -15,6 +15,7 @@
 package attest
 
 import (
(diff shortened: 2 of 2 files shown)


ericchiang:
Got a test log?

brandonweeks:
https://paste.googleplex.com/5686847322718208

Looking into the test failure.

brandonweeks:
Added a test case.

ericchiang (approved):

PR #217, opened 2021-05-19
add ecdsa configuration options

pszal:
Add configuration options for ECDSA key generation.

### DIFF
--- a/attest/application_key.go
+++ b/attest/application_key.go
@@ -71,7 +71,7 @@ type KeyConfig struct {
 	// Algorithm to be used, either RSA or ECDSA.
 	Algorithm Algorithm
 	// Size is used to specify the bit size of the key or elliptic curve. For
-	// example, '256' is used to specify curve P-256."
+	// example, '256' is used to specify curve P-256.
 	Size int
 }
 
--- a/attest/application_key.go
+++ b/attest/application_key.go
@@ -56,9 +56,27 @@ func (s *signer) Public() crypto.PublicKey {
 	return s.pub
 }
 
-// KeyConfig encapsulates parameters for minting keys. This type is defined
-// now (despite being empty) for future interface compatibility.
+// Algorithm indicates an asymmetric algorithm to be used.
+type Algorithm string
+
+// Algorithm types supported.
+const (
+	EC  Algorithm = "EC"
+	RSA Algorithm = "RSA"
+)
+
+// KeyConfig encapsulates parameters for minting keys.
 type KeyConfig struct {
+	// Algorithm to be used, either RSA or EC.
+	Algorithm Algorithm
+	// Size is used to specify the bit size of the key.
+	Size int
+}
+
+// DefaultConfig is used when no other configuration is specified.
+var DefaultConfig = KeyConfig{
+	Algorithm: EC,
+	Size:      256,
 }
 
 // Public returns the public key corresponding to the private key.
--- a/attest/application_key.go
+++ b/attest/application_key.go
@@ -61,21 +61,23 @@ type Algorithm string
 
 // Algorithm types supported.
 const (
-	EC  Algorithm = "EC"
+	ECDSA Algorithm = "ECDSA"
+	// TODO(szp): RSA is not supported yet
 	RSA Algorithm = "RSA"
 )
 
 // KeyConfig encapsulates parameters for minting keys.
 type KeyConfig struct {
-	// Algorithm to be used, either RSA or EC.
+	// Algorithm to be used, either RSA or ECDSA.
 	Algorithm Algorithm
-	// Size is used to specify the bit size of the key.
+	// Size is used to specify the bit size of the key or elliptic curve. For
+	// example, '256' is used to specify curve P-256."
 	Size int
 }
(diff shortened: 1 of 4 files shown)


attest/application_key.go
```diff
@@ -56,9 +56,27 @@ func (s *signer) Public() crypto.PublicKey {
...
 }
 
-// KeyConfig encapsulates parameters for minting keys. This type is defined
-// now (despite being empty) for future interface compatibility.
+// Algorithm indicates an asymmetric algorithm to be used.
+type Algorithm string
+
+// Algorithm types supported.
+const (
+	EC  Algorithm = "EC"
```
ericchiang, line 64: nit: I think this should be "ECDSA"

attest/application_key.go
```diff
@@ -56,9 +56,27 @@ func (s *signer) Public() crypto.PublicKey {
...
+	RSA Algorithm = "RSA"
```
ericchiang, line 65: If we export this variable can you add a comment explaining that it's not supported yet?

attest/application_key.go
```diff
@@ -56,9 +56,27 @@ func (s *signer) Public() crypto.PublicKey {
...
+)
+
+// KeyConfig encapsulates parameters for minting keys.
 type KeyConfig struct {
+	// Algorithm to be used, either RSA or EC.
+	Algorithm Algorithm
+	// Size is used to specify the bit size of the key.
```
ericchiang, line 72: "// Size is used to specify the bit size of the key or elliptic curve. For example, '256' is used to specify curve P-256."

attest/application_key.go
```diff
@@ -56,9 +56,27 @@ func (s *signer) Public() crypto.PublicKey {
...
+	Size int
+}
+
+// DefaultConfig is used when no other configuration is specified.
+var DefaultConfig = KeyConfig{
```
ericchiang, line 77: No real need to export this? You can already specify the DefaultConfig by using nill or an empty key config?

attest/tpm.go
```diff
@@ -354,8 +348,12 @@ func (t *TPM) NewAK(opts *AKConfig) (*AK, error) {
 	return t.tpm.newAK(opts)
 }
 
-// NewKey creates an application key certified by the attestation key.
+// NewKey creates an application key certified by the attestation key. If opts is nil
+// then DefaultConfig is used.
 func (t *TPM) NewKey(ak *AK, opts *KeyConfig) (*Key, error) {
+	if opts == nil {
```
ericchiang, line 354: This breaks the following code?

```
t.NewKey(ak, &attest.KeyConfig{})
```

Probably do something like

```
var defaultConfig = &KeyConfig{
  // ...
}

// ...
if opts == nil {
    opts = &KeyConfig{}
}
if opts.Algorithm == "" && opts.Size == 0 {
    opts = defaultConfig
}
```

attest/wrapped_tpm20.go
```diff
@@ -206,6 +201,68 @@ func (t *wrappedTPM20) newKey(ak *AK, opts *KeyConfig) (*Key, error) {
...
+	}
+
+	return srk, blob, pub, creationData, err
+}
+
+func templateFromConfig(opts *KeyConfig) (tpm2.Public, error) {
+	var tmpl tpm2.Public
+	switch opts.Algorithm {
+	case RSA:
+		return tmpl, fmt.Errorf("not implemented")
```
ericchiang, line 227: nit: more specific error message

```
return tmpl, fmt.Errorf("RSA keys are not implemented")
```

attest/application_key.go
pszal: I wanted to keep it consistent with certtostore but indeed ECDSA is more precise. Done.

attest/application_key.go
pszal: Done. I plan to do an RSA follow-up.

attest/application_key.go
pszal: Done.

attest/application_key.go
pszal: Done.

attest/tpm.go
pszal: Done.

attest/wrapped_tpm20.go
pszal: Done.

attest/application_key.go
```diff
@@ -56,9 +56,29 @@ func (s *signer) Public() crypto.PublicKey {
...
+	// TODO(szp): RSA is not supported yet
+	RSA Algorithm = "RSA"
+)
+
+// KeyConfig encapsulates parameters for minting keys.
 type KeyConfig struct {
+	// Algorithm to be used, either RSA or ECDSA.
+	Algorithm Algorithm
+	// Size is used to specify the bit size of the key or elliptic curve. For
+	// example, '256' is used to specify curve P-256."
```
ericchiang, line 74: nit: looks like a trailing '"' snuck in here. can you remove it?

ericchiang (approved):
one nit, please squash, then :shipit:

PR #222, opened 2021-06-03
Ignore SBAT events in ParseUEFIVariableAuthority

alexmwu:
As part of the Boothole fixes, shim has introduced an SBAT feature https://github.com/rhboot/shim/blob/main/SBAT.md.
SBAT configuration is configured to log to PCR7 using EV_EFI_VARIABLE_AUTHORITY.
https://github.com/rhboot/shim/blob/493bd940e5c6e28e673034687de7adef9529efff/mok.c#L228-L247

This causes issue with ParseUEFIVariableAuthority, as it assumes that an event with type EV_EFI_VARIABLE_AUTHORITY can be parsed as EFI_SIGNATURE_DATA, per section 3.3.4.8 of the TCG PC Client Platform Firmware Profile Specification.

Reproduce: run `TestSecureBootEventLogUbuntu` without the changes. Should get an error like:
```
secureboot_test.go:194: parsing sb state: failed parsing EFI variable authority at event 0: asn1: structure error: tags don't match (16 vs {class:0 tag:24 length:10 isCompound:true}) {optional:false explicit:false application:false private:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false lax:true name:} certificate @2
```


ericchiang (approved):
thanks for the test case!

Issue #202, opened 2021-02-27
Flag to use to avoid installing libtspi-dev.


lihanshang:
Hey Guys
I am running a test and I am seeing 
```
# github.com/google/go-tspi/tspi
/go/pkg/mod/github.com/google/go-tspi@v0.2.1-0.20190423175329-115dea689aad/tspi/context.go:17:27: fatal error: trousers/tss.h: No such file or directory
 // #include <trousers/tss.h>
                           ^
compilation terminated.
```
I am assuming this is the related issue of not installing the libtspi-dev since I am running it on centos. May I know if I dont want TPM12 support what I should run to build? 

I check the relate issues here https://github.com/google/go-attestation/issues/163 but not pretty clear how I should use any flag to make it build.

Thanks!

ericchiang:
You should be able to build without cgo and not include it:

```
CGO_ENABLED=0 go test
```

Does that work?

tracefinder:
Hi @ericchiang,

it works. However, the problem arises when there is another library, which uses cgo. For example, https://github.com/miekg/pkcs11. In this case there is no way to turn off libtspi-dev dependency.

ericchiang:
Ah I see. If you'd like to send a PR to add a build tag (something like "notspi"), we'd be happy to reivew.

ericchiang:
Closed by #232

PR #232, opened 2021-07-29
Add a build tag to turn off TPM12 support and avoid tspi dependency

tracefinder:
The PR provides the ability to turn off TPM12 support to avoid the installation of libtspi-dev (#202).

Right now `go test` fails if you have no libtspi, and `CGO_ENABLED=0 go test` fails too because tests require _setupSimulatedTPM_. _setupSimulatedTPM_ is defined in _attest/attest_simulated_tpm20_test.go_ which needs cgo.

After the PR, it can pass with `go test -tags notspi` (the warning is still there).

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,12 @@ Please note that this is not an official Google product.
 
 The go-attestation package is installable using go get: `go get github.com/google/go-attestation/attest`
 
-Linux users must install `libtspi` and its headers. This can be installed on debian-based systems using: `sudo apt-get install libtspi-dev`.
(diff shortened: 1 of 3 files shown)


tracefinder:
@googlebot I signed it!

ericchiang:
Thanks! Would you mind adding something to the README as well? If not I can take that

https://github.com/google/go-attestation#installation

tracefinder:
Yeah, of course. I just wasn't sure if my description would suit.

ericchiang:
lgtm, thanks!

PR #233, opened 2021-08-06
Fix minor spelling issues in comments

lindi2:
Fix minor spelling issues in comments.

Does the comment in readAllPCRs20 also need updating? It says that the command is repeated up to 8 times but as far as I can understand the loop can call tpm2.ReadPCRs up to 24 times.


josephlr (approved):
Spelling changes look good, thanks for fixing this!

As for `readAllPCRs20`, that comment is wrong. The code, as written, can call ReadPCR at most 24 times. I think the confusion was that ReadPCR can return at most 8 `digests`, not that it needs to be called 8 times.

lindi2:
I pushed one more commit to this PR to update also the misleading comment in readAllPCRs20. Let me know if you'd prefer to have them rebased together instead.

brandonweeks:
Thank you for taking the time to do this!

PR #235, opened 2021-08-23
Handle EFI_ACTION events signalling DMA protection is disabled.

twitchy-jsonp:
Starting to see some use of `ParseSecurebootState()` fail in production due to `event 15: unexpected EFI action event`.

This PR adds support for detecting and verifying that event, so such event logs can be supported as normal.

### DIFF
--- a/attest/pcp_windows.go
+++ b/attest/pcp_windows.go
@@ -26,9 +26,9 @@ import (
 
 	"github.com/google/certificate-transparency-go/x509"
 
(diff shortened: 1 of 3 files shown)


ericchiang:
Got a test case? :P

twitchy-jsonp:
Not one I can check into OSS unfortunately xD

ericchiang (approved):

PR #254, opened 2021-11-26
attest: Make PCRs included in quote configurable

smo4201:
It would be nice, if the PCRs to be included in a Quote could be configured in the low-level API (k *Ak) Quote() method. This would enable attesting only parts of the PCRs (e.g. from PCR17 in a DRTM setup). It does not change the usually used tpm.AttestPlatform() method. I created this pull request to check if this has any chance of getting accepted.

I implemented the change for Windows TPM 1.2 and 2.0 and Linux TPM 2.0. I did not implement the change for the Linux TPM 1.2 as it uses the GetQuote() (https://github.com/google/go-tspi/blob/master/attestation/attestation.go#L149) function of the go-tspi (https://github.com/google/go-tspi) library which also does not support specifying the quote. Currently, I just implemented an error if one tries to specify the PCRs for the Linux TPM 1.2. Of course, I could also propose a PR there to add the PCR selection to the GetQuote() function.  I just wanted to check first, if this PR has any chance of getting accepted. If so, I could of course also sign a CLA.

### DIFF
--- a/attest/attest.go
+++ b/attest/attest.go
@@ -101,7 +101,7 @@ type ak interface {
 	close(tpmBase) error
 	marshal() ([]byte, error)
 	activateCredential(tpm tpmBase, in EncryptedCredential) ([]byte, error)
-	quote(t tpmBase, nonce []byte, alg HashAlg) (*Quote, error)
+	quote(t tpmBase, nonce []byte, alg HashAlg, selectedPCRs []int) (*Quote, error)
 	attestationParameters() AttestationParameters
 	certify(tb tpmBase, handle interface{}) (*CertificationParameters, error)
 }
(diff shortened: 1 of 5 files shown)


ericchiang:
We try to not break API signatures between major versions. If we were going to do this, we'd probably want another API. `QuoteWithPCRs` or something like that.

Is there a reason you'd ever want to quote a subset of the PCRs? Even if you quote all of them, you can generally just throw out the ones you're not interested in.

smo4201:
> 

My goal is to transmit only the necessary information in a generic format (i.e. signed software hashes including the PCR number they were extended to). Without a possibility to select the PCRs that go into the Quote, the PCRs not-of-interest have still to be transmitted to the remote verifier. Otherwise, the verifier cannot aggregate the values to match the aggregated value inside the signed quote. Additionally I would require logic in the verifier to distinguish between the PCRs of interest (where the prover provides the detailed list of signed hashes that were extended) and the PCRs not-of-interest (where the prover just provides the 'random' final PCR value). 

Of course this is possible, I just thought it would be cleaner (and less data transmitted) to specify the PCRs that go into the quote in the first place. 

Of course I understand that you do not want to break API signatures. I could offer to implement a QuoteWithPCRs method if you are willing to add this as an additional API method?

ericchiang:
> Of course I understand that you do not want to break API signatures. I could offer to implement a QuoteWithPCRs method if you are willing to add this as an additional API method?

Yep, maybe even just `QuotePCRs` would work. So, current API is:

```
func (k *AK) Quote(tpm *TPM, nonce []byte, alg HashAlg) (*Quote, error) {}
```

Added method is:

```
// QuotePCRs is like Quote() but allows the caller to select a subset of the PCRs.
func (k *AK) QuotePCRs(tpm *TPM, nonce []byte, alg HashAlg, pcrs []int) (*Quote, error) {}
```

/cc @twitchy-jsonp

You'll need to sign the CLA before we can accept or review the code though https://opensource.google/docs/cla/

smo4201:
I submitted a new PR https://github.com/google/go-attestation/pull/311, now from my corporate account and with a signed CLA. I addressed your feedback and introduced a new method QuotePCRs(). 

I will close this PR now as it is obsoleted by PR#311

PR #264, opened 2022-02-09
Bump go-tpm version

jkl73:
Some breaking changes in the go-tpm

https://github.com/google/go-tpm/pull/237
An addition return parameter in PolicySecret()
https://github.com/google/go-tpm/pull/235
Certify() now returns raw TPMT_SIGNATURE, so no need to pack it.


josephlr (approved):
These changes make sense given the two breaking changes in go-tpm

Issue #92, opened 2019-08-29
attest: Implement EK cert verification, remove verifier/ subpackage


twitchy-jsonp:
Use-case: Verifying an EK certificate chains up to a set of known root CA's and intermediates.

We could leave this to implementers to hack something up using `x509.CertPool`, but there are a few complexities that make that suboptimal:

 - Numerous roots for a few vendors are malformed in ways that Go's x509/asn1 library refuses to parse (but cert-transparency can).
 - Key usages for EK certificate do not map onto the PKI model used for the web, meaning verification using defaults will fail.
 - Unrecognised extensions in the certs cause problems.

Proposal:

 - Make an `EKCertVerifier` structure, that consumes a slice of intermediate & root x509 certificates.

 - Users call `.Verify(der)` and recieve the validated certificate chain, or an error.

WDYT?

twitchy-jsonp:
```go

// EKCertVerifier checks that EK certificates chain up to a set of intermediates
// and root certificates.
type EKCertVerifier struct {
  roots, intermediates *x509.CertPool
}

// NewEKCertVerifier builds an EKCertVerifier object from the given set
// of root & intermediate certificates.
func NewEKCertVerifier(roots, intermediates []byte) (*EKCertVerifier, error) {
  // parse certs --> respective cert pools
}

// Verify checks the EK cert chains up to a known set of root & intermediate
// certificates.
func (v *EKCertVerifier) Verify(cert []byte) (chains [][]*x509.Certificate, err error) {
  // ...
}
```

@ericchiang WDYT?

ericchiang:
This functionality is already available via `"crypto/x509".Certificate.Verify` https://golang.org/pkg/crypto/x509/#Certificate.Verify

I don't see the benefit of adding duplicate API here. Would this be better as documentation?

twitchy-jsonp:
We could get away with documentation if we specify:

 - Youre gonna need to use the cert-transparency x509 package
 - Set `cert.UnhandledCriticalExtensions` to nil in the EKCert before verifying
 - Set `VerifyOptions.KeyUsages` to `[]x509.ExtKeyUsage{x509.ExtKeyUsageAny}`

ericchiang:
sgtm

I think that'd be a good place to start anyway. We should also document why you need to use cert-transparency x509, and why `UnhandledCriticalExtensions` has to be set to nil. Ideally, we'd actually parse those extensions in the future too.

[...]

PR #3, opened 2019-04-03
Update certificate-transparency-go dependency to fix broken reference.

twitchy-jsonp:

PR #269, opened 2022-03-24
Remove certificate-transparency-go dependency 

### LINKED ISSUE #247: Use `crypto/x509` rather than `certificate-transparency-go/x509`
`certificate-transparency-go/x509` relaxes the golang cert parsing and error checking in dangerous ways that should not be used apart from certificate observation use cases.

josephlr:
Fixes #247

Having non-standard x509 structures in the external API makes it very difficult to integrate with other libraries (see https://github.com/google/go-tpm-tools/pull/178). Furthermore, must of the certificate-transparency related differences are not longer needed now that https://github.com/google/go-attestation/issues/92 is fixed.

Looking at the past issues and PRs, I found three places where we needed certificate-transparency-go:
  - https://github.com/google/go-attestation/pull/3
    - No longer needed as `x509.RSAESOAEP` isn't used anymore in this library
  - https://github.com/google/go-attestation/pull/41
    - Using the Standard Library's `asn1` package doesn't seem to cause any issues (all tests pass).
    - The `parseCert` function and use of `asn1.UnmarshalWithParams` no longer exist
  - @twitchy-jsonp's comment https://github.com/google/go-attestation/issues/92#issuecomment-526681092
    -  Overriding `UnhandledCriticalExtensions` and `KeyUsages` can be done with the Standard Library's `x509` package.

The only code changes necessary to get a tests to pass was:
  - Properly use `Fatalf` (instead of `Errorf`)
  - Don't use `x509.IsFatal(err)` as non-fatal errors don't exist in the standard library
  - Don't depend on the exact error code returned by `x509.ParseCertificate` (this wasn't stable anyway)
    - We just unconditionally try the workaround for https://github.com/google/go-attestation/issues/157 if cert parsing fails.

Even if we eventually decide we need to _internally_ use certificate-transparency-go in a few limited places, we should use the standard library types in the external API, `attest.SecurebootState` for example.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,6 +3,11 @@ on:
     branches: [master]
   pull_request:
 
+# Workaround for SHA1 on Go 1.18. There are some kinks to be worked out. See
+# the tracking issue for more info: https://github.com/golang/go/issues/41682
+env:
+  GODEBUG: x509sha1=1
+
 name: Test
 jobs:
   test-linux:
--- a/attest/application_key_test.go
+++ b/attest/application_key_test.go
@@ -28,11 +28,10 @@ import (
 	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/rsa"
+	"crypto/x509"
 	"encoding/asn1"
 	"math/big"
 	"testing"
-
-	"github.com/google/certificate-transparency-go/x509"
 )
 
 func TestSimTPM20KeyCreateAndLoad(t *testing.T) {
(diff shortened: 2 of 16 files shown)


josephlr:
Another use case for certificate-transparency might be: https://github.com/google/go-tspi/pull/24 (i.e. RSAES-OAEP keys). But I'm not sure if go-attestation actually needs this (as it mostly deals with singing keys).

brandonweeks (approved):

josephlr:
@brandonweeks all the tests are passing now.

I used `GODEBUG=x509sha1=1` to get our tests to pass. There is currently a back-port to fix some of the issues around how SHA1 was disable for x509 stuff, so we just reenable it for now. See https://github.com/golang/go/issues/41682 for more info.

PR #274, opened 2022-04-28
Add TPM 2 application key support for Windows

mjg59:
There's currently no support for creating application keys on Windows systems. This patch transitions the Windows key type to specifically refer to attestation keys, and reuses the existing wrapped key support for application keys. This allows the creation of keys in the platform store, while still allowing said keys to be manipulated with existing TPM functionality rather than duplicating it.


hslatman:
Hey @mjg59,

Is there anything I can do to help move this forward? I've been working on getting signing to work with application keys on Windows, building on top of your contributions here, combined with some changes from @brandonweeks. It would be great if my changes can be upstreamed, instead of having to maintain a separate fork.

PR #296, opened 2022-11-04
Preserve error logic in getPrimaryKeyHandle

gerow:
In `wrappedTPM20.getPrimaryKeyHandle()`, preserve any error from the short-circuit `tpm2.ReadPublic()` logic, so that we can return it alongside any failure in `tpm2.CreatePrimary()`


ericchiang (approved):
Thanks!

PR #298, opened 2022-11-15
Truncate digests to the left most bits when signing with ECDSA

mwielgoszewski:
In Golang's ecdsa [signing implementation](https://cs.opensource.google/go/go/+/refs/tags/go1.19.2:src/crypto/ecdsa/ecdsa.go;l=181), digests are truncated to the left most bits to match the bit length of the order of the curve.

If this digest is not truncated, the TPM will return an error `parameter 1, error code 0x15 : structure is the wrong size`.

This PR updates the `signECDSA` function to match what Golang does under the hood.

### DIFF
--- a/attest/wrapped_tpm20.go
+++ b/attest/wrapped_tpm20.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"crypto"
 	"crypto/ecdsa"
+	"crypto/elliptic"
 	"crypto/rsa"
 	"encoding/asn1"
 	"errors"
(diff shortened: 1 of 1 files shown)


brandonweeks:
I believe we're using this code and it works for us. How did you encounter the error?

mwielgoszewski:
I wrote an implementation for [certstore](https://github.com/github/smimesign) for linux that used TPM2 backed keys, and the certstore unit tests caught it:

From https://github.com/github/smimesign/blob/main/certstore/certstore_test.go#L230

```
sha384Digest := sha512.Sum384([]byte("hello"))
sig, err = signer.Sign(rand.Reader, sha384Digest[:], crypto.SHA384)
if err != nil {
	t.Fatal(err)
}
if err = leafEC.Certificate.CheckSignature(x509.ECDSAWithSHA384, []byte("hello"), sig); err != nil {
	t.Fatal(err)
}
```

mjg59:
This breaks P-256 sha256 signing if the leading byte is 0 - the digest loses the leftmost bit (or bits, if there's multiple 0s in a row)

mjg59:
I think the correct fix is probably just to re-add as many leading 0s as are necessary to restore the original digest length?

mjg59:
(Although I don't know if this is correct in the case of a digest longer than the curve length)

PR #300, opened 2022-12-06
Attempt all auth types when reading from EK cert

gerow:


gerow:
Googlers: see b/258300352 for additional context.

brandonweeks:
I don't understand why this change would useful for the policies specified in [TCG EK Credential Profile](https://trustedcomputinggroup.org/wp-content/uploads/TCG_IWG_EKCredentialProfile_v2p4_r3.pdf).

gerow:
> I don't understand why this change would useful for the policies specified in the [TCG EK Credential Profile](https://trustedcomputinggroup.org/wp-content/uploads/TCG_IWG_EKCredentialProfile_v2p4_r3.pdf).

I don't see anywhere in that spec that specifically calls out the appropriate auth handle for reading the EK cert from NVRAM, especially in 2.2.1.9.

brandonweeks:
`TPMA_NV_OWNERREAD = 1` is specified in the "Policy NV", so owner authorization should always be sufficient for reading a compliant EK certificate if the owner password is known.

We know `""` is a valid owner password because we set the owner password. Why would `""` be a valid platform password?

gerow:
@brandonweeks I don't see a situation where the platform handle would work, yes. I'm happy to drop that one. The CrOS vtpm still wants the "self" auth handle since it doesn't have a concept of an owner password.

brandonweeks:
This still doesn't seem right to me. Which policy from [TCG EK Credential Profile](https://trustedcomputinggroup.org/wp-content/uploads/TCG_IWG_EKCredentialProfile_v2p4_r3.pdf) is CrOS using? Does `tpm2_getekcertificate` work with the CrOS vTPM currently?

gerow:
Converting this to a draft for now. I'll look to get the CrOS vtpm working with `tpm2_getekcertificate` first. Once that happens I think I'll be in a better place to see what needs to happen in go-attestation to get it to work (if anything at that point).

PR #310, opened 2023-03-16
Fix Intel EK certificate URL

hslatman:
To download the certificate for an Intel TPM, the base64 padding in the URL needs to be replaced with `%3D`. If it's not replaced, requesting the URL will result in HTTP 403 Forbidden.

### DIFF
--- a/attest/tpm.go
+++ b/attest/tpm.go
@@ -219,7 +219,7 @@ func intelEKURL(ekPub *rsa.PublicKey) string {
 	pubHash.Write(ekPub.N.Bytes())
 	pubHash.Write([]byte{0x1, 0x00, 0x01})
 
-	return intelEKCertServiceURL + base64.URLEncoding.EncodeToString(pubHash.Sum(nil))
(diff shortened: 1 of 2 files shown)


attest/tpm.go
```diff
@@ -219,7 +219,7 @@ func intelEKURL(ekPub *rsa.PublicKey) string {
 	pubHash.Write(ekPub.N.Bytes())
 	pubHash.Write([]byte{0x1, 0x00, 0x01})
 
-	return intelEKCertServiceURL + base64.URLEncoding.EncodeToString(pubHash.Sum(nil))
+	return intelEKCertServiceURL + strings.ReplaceAll(base64.URLEncoding.EncodeToString(pubHash.Sum(nil)), "=", "%3D")
```
brandonweeks, line 222: Could this be a call to [url.QueryEscape](https://pkg.go.dev/net/url#QueryEscape) instead of directly replacing `=`?

hslatman: Fixed in f421e1de0cf130718a2810181a74d9bf1bd3ffdd. Thx! 🙂

brandonweeks (approved):

PR #311, opened 2023-03-28
attest: Make PCRs included in quote configurable

smo4201:
This is an update of https://github.com/google/go-attestation/pull/254, now from my corporate account with a signed CLA. 

The PR introduces an additional function QuotePCRs which is like the existing Quote() but allows configuring the PCRs to be included in a quote. This enables attesting only parts of the PCRs (e.g. from PCR17 in a DRTM setup). 

@ericchiang I addressed your feedback to not introduce any breaking API changes.


ericchiang (approved):
lgtm given this is private. Looks like you need to rebase though

smo4201:
> lgtm given this is private. Looks like you need to rebase though

Thanks and yes, done!

PR #315, opened 2023-04-14
Fix comments referring to `.Serialize()` instead of `.Marshal()`

strideynet:
The comments for `TPM.LoadAK` and `TPM.LoadKey` refer to a `Serialize()` method on `Key` and `AK`. This method does not exist and it appears that `.Marshal()` is the correct method based on the comments there.


ericchiang (approved):

ericchiang:
Thanks!

PR #317, opened 2023-05-09
Improve Windows support

hslatman:
This PR started with the changes from @mjg59 in https://github.com/google/go-attestation/pull/274 and the changes from @brandonweeks in https://github.com/google/go-attestation/compare/master...brandonweeks:go-attestation:acme-device-attest and https://github.com/google/go-attestation/compare/master...brandonweeks:go-attestation:x509ext. I added some functions on top to improve support for Windows.

It would be great if we could get these changes upstreamed. We're currently relying on a renamed fork instead.


ericchiang:
Hey! Thanks so much for sending this and apologies for the delay.

This is a pretty large amount of code. Would it be possible to set up a video call with @brandonweeks and I to chat about this patch? That might be faster than going over github

ericchiang@google.com
bweeks@google.com

ericchiang:
Meeting notes:

* Objectives:
  * High level understanding of what’s being added
  * Trying to get a proposal moved forward for what go-attestation takes on
* https://github.com/google/go-attestation/pull/317
* Smallstep
  * Working on TPM support
  * Using go-attestation 
  * Device attestation spec
  * Needs windows support
* ACME stuff
  * Still a “working group draft” but something we can probably start targeting
* Questions that would help Eric:
  * What are the APIs you want to add
  * What features dependent on other ones
  * How can we help test this?
* Brandon: things are complicated!
  * Internally, we already have windows support working
  * The person who understood this is now at Tailscale (https://github.com/twitchy-jsonp)
  * We currently use a hack to creat the AK because creating AKs directly is hard
  * Generate an application key, causes the AK to be generated automatically
  * For reasons, there’s issues with generating AK directly due to go-tpm missing delegation support
* Herman:
  * Not familiar with those restrictions
  * Took matthew garrett’s stuff
* Brandon:
  * Generating an AK with imprint
  * Windows uses way more the authorizations of TPMs
  * There are owners passwords and delegated authorization keys in the registry
  * Technically possible to open up the interaction with Windows directly
  * Workaround: generate the attestation key with ncrypt, causes Windows to generate the AK for you
  * Example of where this can have a limitation:
    * you can’t specify the template for Windows, even though you can for Linux
* Herman:
  * We’re running as Admin, does that change this?
* Brandon:
  * You can send thing commands but can’t setup the other delegated authorization credentials
* Eric:
  * Let’s just add APIs and don’t worry about where it’s implemented on Windows or Linux
  * Can we “just” add delegated authorization to go-tpm?
* Brandon: was told 4 years ago this was hard
  * go-tpm is getting a major upgrade
  * Maybe a good time to talk to the other go-tpm teams?
* Eric:
  * maybe we just provide examples for go-tpm?
  * E.g. if users are already using go-tpm. Maybe we shouldn't bother wrapping it with go-attestation?
  * Could we provide examples, or small packages to go-tpm that works with that package rather that wrapping it?

hslatman:
@ericchiang just saw @mjg59 pushing the changes from this PR to his branch, so I think that's what's going to be tested in full now.

Short summary of what's being changed / added:

- Windows application key support from @mjg59
- ACME `device-attest-01` changes from @brandonweeks: `QualifyingData` in `KeyConfig` and related logic, mainly.
- Fixes to make the initial Windows changes work. 
- Fixes to make `device-attest-01` work on Windows to from my side

Then the additions, all dependent on the Windows changes:
- A new `Name` property in `AKConfig` and logic to control the name of the AK generated on Windows. Empty by default, results in the existing behavior. Makes it a bit simpler to reference existing AKs on Windows.
- A new `Name` property in `KeyConfig`. Same as for the AK.
- Added `Blobs` method to `AK`. Already existed for `Key`.
- Added `DeleteAK` and `DeleteKey` methods.

Let me know if you would like references to the actual lines. Changes are currently spread over multiple commits; I can clean those up and get some nicer commits in if that's preferred.

mjg59:
I have a couple of cleanups and one patch that fixes ECDSA signing for me built on top of Herman's work - I'll push those separately for review, but 4 patches gets us to the point where application keys can be generated and used for signing. I think it makes sense to land that functionality and then move forward on the additional work? (I just successfully got an application cert issued and stored in the Windows cert store backed by a TPM key)

mjg59:
Updated https://github.com/google/go-attestation/pull/274 with the minimal set of patches that lands useful new functionality. Selfishly I think this entirely deals with my use case, but I think it's a reviewable chunk of code that we can build on for the rest of this :)

hslatman:
I'll start with the current state of #274 and move my additions/changes on top of that.

@ericchiang multiple PRs makes sense; I'll try to make that work. Some changes were fairly small though, so I might combine some again.

PR #318, opened 2023-05-15
Parse TCG_PCR_EVENT2 structures with an eventSize of 0

xose:
Similar to #212, support parsing TPM2 events with an empty data field.

### DIFF
--- a/attest/eventlog_test.go
+++ b/attest/eventlog_test.go
@@ -174,6 +174,43 @@ func TestParseEventLogEventSizeZero(t *testing.T) {
 	}
 }
 
+func TestParseEventLog2EventSizeZero(t *testing.T) {
+	data := []byte{
+		// PCR index
+		0x0, 0x0, 0x0, 0x0,
+
+		// type
+		0x7, 0x0, 0x0, 0x0,
+
+		// number of digests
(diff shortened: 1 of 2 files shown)


ericchiang (approved):
Thanks!

attest/eventlog_test.go
```diff
@@ -174,6 +174,43 @@ func TestParseEventLogEventSizeZero(t *testing.T) {
...
+
+		// Event size (0 B)
+		0x0, 0x0, 0x0, 0x0,
+
+		// no "event data"
+	}
+
+	specID := &specIDEvent{
+		algs: []specAlgSize{
+			specAlgSize{ID: uint16(tpm2.AlgSHA256), Size: 32},
```
ericchiang, line 205: I think lint is complaining about the redundant use of specAlgSize. You can omit it in this case. For example:

```
$ cat hi.go 
package a

type b struct {
	c int
}

var d = []b{
	b{c: 1},
	b{c: 2},
}
$ gofmt -s hi.go 
package a

type b struct {
	c int
}

var d = []b{
	{c: 1},
	{c: 2},
}
```

Mind fixing that?

xose: Done.

ericchiang (approved):

Issue #320, opened 2023-05-18
Handle "AttestedCertifyInfo" in ActivationParameter checks for TPM2.0 Challenges


juanvallejo:
ActivationParameters currently only supports handling the `tpm2.TagAttestCreation`. Support for `tpm2.TagAttestCertify` is currently missing.

The context behind this change is to add the ability to create an Activation Challenge for a TPM, using AttestedCertify info instead of the currently-supported AttestedCreation info.

None of the current checks against ActivationParameters would change as a result.
The main consequence of this would be that the AIK value used in generating an activation challenge would be [derived from the name digest of an AttesttedCertify structure instead of an AttestedCreation structure](https://github.com/google/go-attestation/blob/master/attest/activation.go#LL242C50-L242C86).

cc @ericchiang

ericchiang:
Does this require a change to the AK the way we generate the AK, or is this specifically only related to the challenge?

Would that look like adding a field to opt into attest certify rather than attest creation? E.g. if we had something like

```go
type AttestationType int

const (
    AttestCreation AttestationType = iota // Default value
    AttestCertify
)
```

Then let you specify that field in ActivationParameters and CertificationParameters, would that work?

https://pkg.go.dev/github.com/google/go-attestation/attest#ActivationParameters
https://pkg.go.dev/github.com/google/go-attestation/attest#CertificationParameters

juanvallejo:
> Does this require a change to the AK the way we generate the AK, or is this specifically only related to the challenge?
This is only specifically to the challenge.

> Would that look like adding a field to opt into attest certify rather than attest creation? E.g. if we had something like
When we parse attestation data using [tpm2.DecodeAttestationData](https://github.com/google/go-tpm/blob/master/tpm2/structures.go#L667), the bytes we initially receive for attestation data should already have a [tag that matches one of the ones already present in the tpm2 library](https://github.com/google/go-tpm/blob/master/tpm2/constants.go#L402) to determine whether we're dealing with AttestedCertify data or AttestedCreation data. So no further changes needed, especially given that the library already recognizes both tag-types we intend to handle here.

ericchiang:
Cool. I will note that one of the explicit goals of this package is to avoid the need to use go-tpm2 directly. So if we add support for verifying a particular challenge type, we should also add support for generating it.

I don't think it makes sense for us to only support verification.

PR #322, opened 2023-05-18
Add support for generating TPM2.0 challenges using AttestedCertifyInfos

juanvallejo:
Fixes: #320.

Adds support for generating an activation challenge using CertificationParameters.
Achieves symmetry with challenge-generation in AttestationParameters, in order to provide a challenge to a TPM to activate a TPM-certified key.

`attest.Activation` currently supports verifying and generating a challenge given attestationData, an EK, an AK, and a signature. In the attestationData, the CreationInfo field is used to further validate and create the resulting challenge.
In this change, `attest.Certification` will now support generating a challenge given attestationData, an EK, a TPM-certified public key, and a signature, in addition to an AK used to verify the certification of the provided public key we are generating an activation challenge for.

cc @ericchiang

### DIFF
--- a/attest/certification.go
+++ b/attest/certification.go
@@ -17,12 +17,14 @@ package attest
 import (
 	"bytes"
 	"crypto"
+	"crypto/rand"
 	"crypto/rsa"
 	"errors"
 	"fmt"
 	"io"
 
 	"github.com/google/go-tpm/tpm2"
+	"github.com/google/go-tpm/tpm2/credactivation"
 	"github.com/google/go-tpm/tpmutil"
 )
 
@@ -60,6 +62,38 @@ type VerifyOpts struct {
 	// Hash is the hash function used for signature verification. It can be
 	// extracted from the properties of the certifying key.
 	Hash crypto.Hash
+}
+
+// ActivateOpts specifies options for the key certification's challenge generation.
+type ActivateOpts struct {
+	// EK, the endorsement key, describes an asymmetric key whose
+	// private key is permanently bound to the TPM.
+	//
+	// Activation will verify that the provided EK is held on the same
+	// TPM as the key we're certifying. However, it is the caller's responsibility to
+	// ensure the EK they provide corresponds to the the device which
+	// they are trying to associate the certified key with.
+	EK crypto.PublicKey
+	// VerifierKeyNameDigest is the name digest of the public key we're using to
+	// verify the certification of the tpm-generated key being activated.
+	// The verifier key (usually the AK) that owns this digest should be the same
+	// key used in VerifyOpts.Public.
+	// Use tpm2.Public.Name() to produce the digest for a provided key.
+	VerifierKeyNameDigest *tpm2.HashValue
+}
+
+// NewActivateOpts creates options for use in generating an activation challenge for a certified key.
+// The computed hash is the name digest of the public key used to verify the certification of our key.
+func NewActivateOpts(verifierPubKey tpm2.Public, ek crypto.PublicKey) (*ActivateOpts, error) {
+	pubName, err := verifierPubKey.Name()
+	if err != nil {
+		return nil, fmt.Errorf("unable to resolve a tpm2.Public Name struct from the given public key struct: %v", err)
+	}
+
+	return &ActivateOpts{
+		EK:                    ek,
+		VerifierKeyNameDigest: pubName.Digest,
+	}, nil
 }
 
 // Verify verifies the TPM2-produced certification parameters checking whether:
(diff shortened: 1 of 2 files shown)


ericchiang (commented):
I think I implied this in https://github.com/google/go-attestation/issues/320, but I don't think it makes sense for go-attestation to support verifying challenges that it can't generate. Is there a reason there's not an associated change to ActivationParameters.Generate()?

Also would it be possible to add a test?

juanvallejo:
> I think I implied this in #320, but I don't think it makes sense for go-attestation to support verifying challenges that it can't generate. 

At least for use-cases I've come across so far, attestation data can be generated by the TPM in different types ([Create & Certify, for example](https://github.com/google/go-tpm/blob/master/tpm2/structures.go#L694)). For now, the https://github.com/google/go-tpm repo supports these two attestation data types, which lends itself to the suggestion to add similar support here to verify attestation data of these two types.

> Is there a reason there's not an associated change to ActivationParameters.Generate()?

There is an associated change in that function, the change is under the specific helper to generate the challenge for TPM 2.0 here: https://github.com/google/go-attestation/pull/322/files#diff-f615933412436eaa513180445920faeaf2cb8cd9ed620eac35f3eaa4012933fcR284

> Also would it be possible to add a test?

Definitely, will add a test.

juanvallejo:
@ericchiang Per the followup offline conversation, I have revised this patch to complete generating an activation challenge for [CertificationParameters](https://github.com/google/go-attestation/blob/master/attest/certification.go#L41).

Initially, my idea was to patch activation.go to support generating an activation challenge for a tpm-created AK as well as 
a tpm-certified public key (our specific use-case).

However, given that CertificationParameters exist independently of `ActivationParameters` and its nested AK `AttestationParameters`, I have decided to add support for generating a similar activation challenge to `CertificationParameters` instead.

attest/certification.go
```diff
@@ -62,6 +64,20 @@ type VerifyOpts struct {
...
+type ActivateOpts struct {
+	// EK, the endorsement key, describes an asymmetric key whose
+	// private key is permanently bound to the TPM.
+	//
+	// Activation will verify that the provided EK is held on the same
+	// TPM as the key we're certifying. However, it is the caller's responsibility to
+	// ensure the EK they provide corresponds to the the device which
+	// they are trying to associate the certified key with.
+	EK crypto.PublicKey
+	// Hash is the hash structure used as
```
ericchiang, line 77: Can you complete the comment here?

Is this required? How does someone compute this value? Does this imply ActivateOpts only supports tpm2?

attest/certification_test.go
```diff
@@ -274,3 +276,72 @@ func testKeyCertification(t *testing.T, tpm *TPM) {
...
+		EK:   eks[0].Public,
+		Hash: pubName.Digest,
+	}
+
+	sk, err := tpm.NewKey(ak, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	rand := mathrand.New(mathrand.NewSource(123456))
```
ericchiang, line 331: why does this need to be a consistent value and not use a regular rand.Reader? Can you add a comment at the very least?

attest/certification_test.go
```diff
@@ -274,3 +276,72 @@ func testKeyCertification(t *testing.T, tpm *TPM) {
...
 	}
 }
+
+func TestKeyActivationTPM20(t *testing.T) {
+	sim, tpm := setupSimulatedTPM(t)
+	defer sim.Close()
+
+	ak, err := tpm.NewAK(nil)
+	if err != nil {
+		t.Fatal(err)
```
ericchiang, line 286: nit: please include good error messages. it really helps triage failures

t.Fatalf("creating new tpm: %v", err)

(this comment applies to the rest of this function)

juanvallejo:
cc @sblaes

attest/certification_test.go
juanvallejo: Will leave as a regular rand.Reader as it does not have to be a fixed value

attest/certification.go
juanvallejo: Renamed this value to be more descriptive. This value is the name digest of the key we're using to verify certification of the target key being activated. I've added a comment to explain how to derive this value and added a helper to construct an ActivateOpts struct

juanvallejo: > Does this imply ActivateOpts only supports tpm2?

That is correct, [though based on this comment on the existing CertificationParameters](https://github.com/google/go-attestation/blob/master/attest/certification.go#L40), this file currently only supports tpm2

attest/certification_test.go
juanvallejo: Sounds good, added more descriptive errors throughout this function

attest/certification.go
```diff
@@ -157,6 +176,65 @@ func (p *CertificationParameters) Verify(opts VerifyOpts) error {
 	return nil
 }
 
+// NewActivateOpts creates options for use in generating an activation challenge for a certified key.
+// The computed hash is the name digest of the public key used to verify the certification of our key.
+func (p *CertificationParameters) NewActivateOpts(verifierPubKey tpm2.Public, ek crypto.PublicKey) (*ActivateOpts, error) {
```
ericchiang, line 181: This method doesn't seem to use its receiver?

Maybe have this be documentation above the VerifierKeyNameDigest field in ActivateOpts?

```
// Use tpm2.Public.Name() to produce the digest for a provided key. 
```

juanvallejo:
@ericchiang friendly ping on latest changes, wondering if this PR looks good to go?

ericchiang (approved):
Sorry, thanks for the ping. lgtm!

PR #331, opened 2023-06-13
use legacy tpm2 at its new path

chrisfenner:
As part of releasing go-tpm 0.9.0, we need to update go-tpm-tools to use go-tpm/tpm2 at its new legacy path (go-tpm/legacy/tpm2). However, go-tpm-tools also depends on this package, which depends back on go-tpm-tools.

This change updates go-attestation to depend on go-tpm @ 511507721cb172faf1149ab9e38ff7881eb1b776 (which adds tpmdirect but doesn't yet delete the Startup adapter), and use legacy/tpm2 instead of tpm2. This will allow a future change to go-tpm-tools to do the same.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
   test-linux:
     strategy:
       matrix:
-        go-version: [1.19.x, 1.20.x]
+        go-version: [1.20.x]
     runs-on: ubuntu-latest
     steps:
     - name: Install Go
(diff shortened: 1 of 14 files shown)


chrisfenner:
CI test is reporting:

```
"crypto/ecdsa" imported but not used (typecheck)
```

Not sure to which file this error refers. `ecdsa` seems to be used in all the files in this project that reference it...

brandonweeks (approved):

PR #332, opened 2023-06-14
Add EK as a field to AK struct.

zhsh:
The change is a no-op for existing clients, and it will simplify adding the support for ECC EKs. The activation code no longer makes assumptions about EK's type and handle (i.e. RSA and 0x81010001), and instead relies on TPM.EKs() to provide the EK's details.

### DIFF
--- a/attest/attest.go
+++ b/attest/attest.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/google/go-tpm/legacy/tpm2"
 	"github.com/google/go-tpm/tpm"
+	"github.com/google/go-tpm/tpmutil"
 )
 
 // TPMVersion is used to configure a preference in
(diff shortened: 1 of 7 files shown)


zhsh:
The current way to create an AK:
```
eks, err := tpm.EKs()
if err != nil {
    // handle error
}
ek := eks[0]

akConfig := &attest.AKConfig{}
ak, err := tpm.NewAK(akConfig)
```

The new way to create an AK:

```
eks, err := tpm.EKs()
if err != nil {
    // handle error
}
ek := eks[0]

akConfig := &attest.AKConfig{EK: &ek}
ak, err := tpm.NewAK(akConfig)
```

zhsh:
I can't quite understand what the lint check fails: https://github.com/google/go-attestation/actions/runs/5263024313
The check complains about "crypto/ecdsa" imported but not used in attest/attest-tool/attest-tool.go, while it is clearly used there (also the PR does not change the file).

brandonweeks (approved):

PR #338, opened 2023-06-23
attest: Add AK signing interface

smo4201:
Restricted TPM keys, such as AKs, are usually used to sign objects generated within the TPM, such as quotes. However, they can also be used for signing data that was not generated within the TPM, if the data does not start with TPM_GENERATED_VALUE. This requires hashing the data to-be-signed by the TPM and retrieving a validation ticket together with the digest from the TPM. 

It can be useful, e.g., to generate a self-signed Certificate Signing Request for the AK and then retrieve AK certificates from CAs (which are able to perform credential activation) via standard protocols such as Enrollment over Secure Transport (EST). EST mandates that the CSR is self-signed with the corresponding private key (RFC 7030, Section 4.2.1), so this is imho the only possibility to be EST compliant. 

This pull request therefore adds a sign() method to the AK interface for signing arbitrary objects. Furthermore, it adds a hash() method to the TPM interface, as the data to-be-signed by a restricted key must be hashed within the TPM to retrieve a TPM validation ticket. The existing internal sign() method of the wrappedKey20 implementation was extended with a validation ticket parameter to be able to prove that the data was hashed within the TPM. 

Note: currently, I only implemented this additional functionality for the Linux TPM2.0. The TPM1.2 and Windows TPM throw a "not implemented" error if the new methods are called.

### DIFF
--- a/attest/application_key.go
+++ b/attest/application_key.go
@@ -26,7 +26,7 @@ type key interface {
 	close(tpmBase) error
 	marshal() ([]byte, error)
 	certificationParameters() CertificationParameters
-	sign(tpmBase, []byte, crypto.PublicKey, crypto.SignerOpts) ([]byte, error)
+	sign(tpmBase, []byte, crypto.PublicKey, crypto.SignerOpts, any) ([]byte, error)
 	decrypt(tpmBase, []byte) ([]byte, error)
 	blobs() ([]byte, []byte, error)
 }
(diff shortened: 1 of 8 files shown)


JasonXJ:
Hi @smo4201, we just merged in some AK signing code (#434) that we have been using internally. Can you see if it satisfys your need? At the meantime, I am going to close this for now.

PR #343, opened 2023-07-17
Update Windows application key support

hslatman:
Catching up to the earlier work on https://github.com/google/go-attestation/pull/274 and https://github.com/google/go-attestation/pull/317. This is the first of several PRs to get back to the state of https://github.com/google/go-attestation/pull/317, but in more manageable chunks.

There's been a bit of activity in terms of (new) function signatures in this repository lately. This PR takes @mjg59's changes and additions for Windows application key support and I applied the required changes to make it compile with the new function signatures.

@ericchiang you asked about the tests on Windows: https://github.com/google/go-attestation/pull/317#issuecomment-1561423479. It's currently only a `go build`; nothing seems to be actually tested outside of the code compiling successfully. I might give that a stab in the upcoming PRs.

P.S.: I added the `-f` to the macOS CI to get around an issue with the linked library already existing: https://github.com/hslatman/go-attestation/actions/runs/5575815478/jobs/10186236329#step:5:8. I'm not sure why that resulted in an error in that run and did not before, outside of a slightly new image being used, but I think the override is OK.

cc @brandonweeks


hslatman:
If I (re)open the PR, is there some indication that it'll be reviewed within a reasonable time? 

As far as I know, the changes from @mjg59 weren't merged so far, and I don't think a functional equivalent PR was merged either, so Instill think they're useful additions.

PR #357, opened 2023-11-09
Don't drop leading zeroes when performing generic ecdsa signing

mwielgoszewski:
As @mjg59 pointed out in [#298](https://github.com/google/go-attestation/pull/298#issuecomment-1802877908), this implementation breaks signing when the digest leads with a zero byte. 

To fix this, call ret.FillBytes() instead of ret.Bytes() to preserve leading zeroes that may have been dropped when converting the digest to an integer.


mjg59:
Verified this fixes our issue

brandonweeks:
Thank you testing Matthew!

brandonweeks (approved):

Issue #366, opened 2024-03-25
 MarshalSubjectAltName should  support specificing if the extension is critical.


liamjm:
x509/x509ext.go MarshalSubjectAltName explicitly sets the extension to non-critical.

From the [specification of AK Certs (TPM 2.0 Keys for Device Identity and Attestation)](https://trustedcomputinggroup.org/wp-content/uploads/TPM-2p0-Keys-for-Device-Identity-and-Attestation_v1_r12_pub10082021.pdf), section 8, 

> The subjectAltName extension is a standard X.509v3 extension. The X.509 certificate profile presented in RFC5280 specifies the subjectAltName extension for allowing the binding of additional identities to the subject of the
> certificate. In an IDevID/IAK certificate, the subjectAltName SHOULD NOT be critical, as this could introduce
> interoperability concerns.
> 

and 

> Note: If the LDevID/LAK creator decides to leave the subject name field empty, the subjectAltName extension
> MUST be critical in accordance with RFC 5280 [4].
> 

From the [EK Spec (TCG EK Credential Profile)](https://trustedcomputinggroup.org/wp-content/uploads/TCG-EK-Credential-Profile-V-2.5-R2_published.pdf), section 3.2.9 Subject Alternative Name

> In accordance with RFC 5280 [12], this extension MUST be critical if subject is empty and
> SHOULD be non-critical if subject is non-empty
> 

So MarshalSubjectAltName() should allow the caller to specify if extension is critical.

PR #367, opened 2024-03-25
Adjust MarshalSubjectAltName to allow SANs to be critical

liamjm:
For both AKs and EKs the Subject Alt Name extension may be critical, or non-critical, based on the presence of the subject field.

So we need to expose control over this criticality to the caller, as detailed more in #366 .

This PR adds a new method MarshalSubjectAltNameWithCriticality that accepts a bool to control the criticality. The existing method (MarshalSubjectAltName) that specifies the extension non-critical is left untouched to maintain compatibility.

### DIFF
--- a/x509/x509ext.go
+++ b/x509/x509ext.go
@@ -149,8 +149,9 @@ func forEachSAN(extension []byte, callback func(ext asn1.RawValue) error) error 
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


brandonweeks (changes requested):
looks good, can you please squash the commits?

liamjm:
Squashing done, and update from master branch.

brandonweeks (approved):

PR #421, opened 2025-05-05
Make ActivationParameters.CheckAKParameters() public

JasonXJ:
This allows the user to check the AK parameter even if it does not need to generate the encrypted credential.


brandonweeks:
LGTM but needs rebase + lint fix.

JasonXJ:
Thanks, Brandon. It should be fixed now.

PR #429, opened 2025-06-17
Support ECDSA verification for attribute certificates

limansa:


liamjm:
So typically I would want tests for any new code, but alas there are not any tests for this code currently. And there is not any method to create one of these attribute certs. So I don't really think I can make this demand. 
And in any case.... I am writing a PR for adding a method to create attributecerts, including support for ECDSA. The changes in this PR are pretty similar to what I'm working up, so I am happy with this PR, regardless of the lack of testing.

Thanks for your contribution.

limansa:
Thanks to Liam Murphy for the review and approval.

PR #444, opened 2025-09-10
feat: support quote for ECDSA based AK

### LINKED ISSUE #443: Support Quote for ECDSA based AK
## Motivation

Improve features support for ECDSA based AK. 

After some experiments, I've discovered that `Quote` features (i.e. generation & validation) wasn't supported yet.

#### Generation

> [!NOTE]
> Only RSA signature is supported

https://github.com/google/go-attestation/blob/4f3c3b0fe5706286182530cd798be833ad0c4a74/attest/tpm.go#L334

#### Validation

> [!NOTE]
> Only RSA public key is supported

https://github.com/google/go-attestation/blob/4f3c3b0fe5706286182530cd798be833ad0c4a74/attest/eventlog.go#L261-L273

---

Knowing that `wrappedKey20` and `windowsKey20` both internally call `quote20` function (i.e. *tpm.go*) the feature will be available for each OS 💫 .

loicsikidi:
fixes #443

### DIFF
--- a/attest/eventlog.go
+++ b/attest/eventlog.go
@@ -17,6 +17,7 @@ package attest
 import (
 	"bytes"
 	"crypto"
+	"crypto/ecdsa"
 	"crypto/rsa"
 	"crypto/sha1"
 	"crypto/sha256"
@@ -257,6 +258,7 @@ func (a *AKPublic) validateQuote(quote Quote, pcrs []PCR, nonce []byte) error {
 
 	sigHash := a.Hash.New()
 	sigHash.Write(quote.Quote)
+	digest := sigHash.Sum(nil)
 
 	switch pub := a.Public.(type) {
 	case *rsa.PublicKey:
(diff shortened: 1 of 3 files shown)


attest/eventlog.go
```diff
@@ -257,18 +258,19 @@ func (a *AKPublic) validateQuote(quote Quote, pcrs []PCR, nonce []byte) error {
 
 	sigHash := a.Hash.New()
 	sigHash.Write(quote.Quote)
+	digest := sigHash.Sum(nil)
 
 	switch pub := a.Public.(type) {
 	case *rsa.PublicKey:
-		if sig.RSA == nil {
-			return fmt.Errorf("rsa public key provided for ec signature")
-		}
```
u1f35c, line 265: Why are we dropping this? Really we should end up with a counterpart for the ECC side to make sure we have an ECC signature for an ECC key.

attest/eventlog.go
```diff
@@ -257,18 +258,19 @@ func (a *AKPublic) validateQuote(quote Quote, pcrs []PCR, nonce []byte) error {
...
 		sigBytes := []byte(sig.RSA.Signature)
-		if err := rsa.VerifyPKCS1v15(pub, a.Hash, sigHash.Sum(nil), sigBytes); err != nil {
+		if err := rsa.VerifyPKCS1v15(pub, a.Hash, digest, sigBytes); err != nil {
+			return fmt.Errorf("invalid quote signature: %v", err)
+		}
+	case *ecdsa.PublicKey:
+		if !ecdsa.Verify(pub, digest, sig.ECC.R, sig.ECC.S) {
 			return fmt.Errorf("invalid quote signature: %v", err)
```
u1f35c, line 271: There's no err here to report; we get a `bool` back from Verify. Just drop the ": %v"

u1f35c (changes requested):
I've just ended up implementing something very similar locally, and would love to see ECC AK support land.

liamjm:
LGTM.

Thanks

Issue #447, opened 2025-10-02
Does this library have an active maintainer?


abovestratosphere:
Currently trying to scope out if this library is suitable for a project or if using go-tpm-tools or go-tpm directly is more suitable. This library exposes a nice abstraction over the ActivateCredential flow which fits my needs (and the other two don't seem to have that), but the lack of recent commits concerns me. The README of the library says it's under active development, but that does not seem to be the case. Is this library abandoned? Are people encouraged to use go-tpm or go-tpm-tools instead?

brandonweeks:
This library is used in production at Google and several other large companies. Development has slowed because the library achieved the original design goals, not because it has been abandoned. The README should probably be updated to reflect that instead of using the "under active development" language.

One caveat is that support for Windows is still relatively immature compared to support for Linux.

Disclaimer: I no longer work at Google and am not speaking on their behalf.

abovestratosphere:
Thanks for the clarification.

PR #449, opened 2025-10-08
Update README.md

liamjm:
update README.md text to reflect that this project has reached initial maturity.

Motivation: #447


liamjm:
@brandonweeks PTAL.

PR #477, opened 2025-12-03
Fix a bug that KeyConfig.QualifyingData is ignored

JasonXJ:
by tpm.NewKey...()  when the algorithm and size are not set.


liamjm (approved):
Thanks!

PR #490, opened 2026-03-20
fix: reject oversized SIPA event headers

jlory:
I noticed very high memory usage while fuzzing windows SIPA events, this caused the test to crash.

```readLoadedModuleAggregation``` and ```readELAMAggregation``` initialized an io.LimitedReader directly from the untrusted ```header.Size``` field without first checking it against the available data. An attacker-controlled event log could create a header of size up to ~4GB causing a large allocation before any read fails.

About the fix, the code is now returning an error, I think its better to fail early than try to move on and use the smallest value ( header size vs len of available bytes in the reader), thoughts?

### DIFF
--- a/attest/win_events_test.go
+++ b/attest/win_events_test.go
@@ -15,6 +15,8 @@
 package attest
 
 import (
+	"crypto/sha256"
+	"encoding/binary"
 	"encoding/json"
 	"os"
 	"testing"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jlory** [comment]
@brandonweeks @liamjm Some PR have been sitting for some time, do you know who can look at it? Thank you,

**Comment by jlory** [comment]
Any updates @liamjm , I think you forgot this one ;)

File: attest/win_events_test.go [hunk 1]
```diff
@@ -23,111 +25,145 @@ import (
...
+		}
+	})
+
+	t.Run("oversized ELAM aggregation header", func(t *testing.T) {
+		if _, err := ParseWinEvents([]Event{oversizedSIPAEvent(t, elamAggregation)}); err == nil {
+			t.Error("expected error for oversized ELAM aggregation header, got nil")
+		}
+	})
+
+	t.Run("golden", func(t *testing.T) {
```
liamjm: Why change the successful case like this?

I would prefer and it would be easier to read IMHO if the failure cases were moved to their own test method (eg TestParseWinEvents_Failures)

liamjm (changes requested):

liamjm:
Sorry about the delay, this one slipped through.

attest/win_events_test.go
jlory: Sorry for the unintended change, I reverted TestParseWinEvents to its original structure and moved the failure cases to a new TestParseWinEvents_Failures function.

liamjm (approved):

PR #492, opened 2026-04-11
attest: return error when ECC public key parsing fails in verifyECDSASignature

kuranikaran:
`verifyECDSASignature` in `attest/activation.go` returns `nil` when
`pub.Key()` fails, silently skipping ECDSA signature verification
for malformed ECC keys. This allows `ActivationParameters.Generate()`
to succeed without validating `CreateSignature`.

This is the same class of issue as CVE-2022-0317 but in the ECC
code path. A malformed ECC TPMT_PUBLIC causes `Key()` to return an
error, which is silently swallowed, so the attestation flow proceeds
without ever checking the AK's creation signature.

The fix returns the error instead of `nil`.

### DIFF
--- a/attest/activation.go
+++ b/attest/activation.go
@@ -180,7 +180,7 @@ func verifyRSASignature(pub tpm2.Public, p *ActivationParameters) error {
 func verifyECDSASignature(pub tpm2.Public, p *ActivationParameters) error {
 	key, err := pub.Key()
 	if err != nil {
-		return nil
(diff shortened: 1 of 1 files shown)


jlory:
This is the same fix as another PR: https://github.com/google/go-attestation/pull/491

liamjm:
Thanks for the submission. As said in #issuecomment-4244824408, this is a dupe of #491.

PR #380, opened 2024-07-11
Add support for AMD EK certificates

Krishap-s:
This PR adds support for retrieving EK certificates from AMD remotely. This has been tested on my AMD Ryzen  4500u fTPM.

PR #495, opened 2026-04-15
Bug fix: use manufacturer string in ekCertURL function

liamjm:
ekCertURL compares manufacturer against URL constants (e.g. intelEKCertServiceURL) instead of name constants (e.g. manufacturerIntel), making EK cert fetching fail.

This code was added in #380


liamjm:
@JasonXJ

JasonXJ (approved):

PR #496, opened 2026-04-22
test: add ECDSA activation regression coverage

1seal:
## what changed

this adds regression coverage for the ECDSA AK verification path in `attest/activation.go`.

the new tests verify that:

- `CheckAKParameters()` rejects ECC AK inputs when `pub.Key()` fails to extract an ECDSA public key
- `CheckAKParameters()` rejects invalid ECDSA signatures on a valid curve

## why

the fail-open in `verifyECDSASignature()` was fixed in `5e9eb5d`, but there was no focused regression test that locks that behavior in place.

these tests make sure malformed ECC parameters cannot silently bypass attestation signature verification again.

## validation

- `go test -v ./attest -run 'Test(CheckAKParametersRejectsECCPublicKeyParseFailures|CheckAKParametersRejectsInvalidECDSASignature|Activation|ECCActivation)$'`
- `go test ./attest`


liamjm:
Thanks for this.
Is there any reason why this can't be done in activation_test.go? This PR improves coverage and doesn't necessarily need the 'regression' concept.

1seal:
@liamjm done, thanks. moved these into activation_test.go so the coverage lives with the rest of the activation tests.

validation:
go test -v ./attest -run 'Test(CheckAKParametersRejectsECCPublicKeyParseFailures|CheckAKParametersRejectsInvalidECDSASignature|Activation|ECCActivation)$'
go test ./attest

liamjm (approved):

PR #500, opened 2026-05-06
attest/internal: reject EFI_SIGNATURE_LIST with oversized SignatureSize to prevent OOM

evilgensec:
## Summary

`parseEfiSignatureList` in `attest/internal/events.go` validates `SignatureListSize` against `maxDataLen` (1 MiB) and prevents underflow in `SignatureSize`, but places **no upper bound** on `SignatureSize`. This allows a crafted event log to trigger a ~4 GiB heap allocation before `binary.Read` returns an error.

### Attack scenario

In remote-attestation deployments a verifier calls `ParseEventLog` + `ParseSecurebootState` on a log received from an (untrusted) attester. A malicious attester sends an event log containing an `EFIVariableDriverConfig` event (e.g. for `PK`, `KEK`, `db`, or `dbx`) whose `EFI_SIGNATURE_LIST` has:

| Field | Value |
|---|---|
| `SignatureType` | `certX509SigGUID` or `hashSHA256SigGUID` |
| `SignatureListSize` | `efiSignatureListHeaderSize + 1` (29 bytes) — passes the `maxDataLen` check |
| `SignatureSize` | `0xFFFFFFFF` — passes the `>= efiGUIDSize` check |

The loop condition `uint32(sigOffset) < SignatureListSize - efiSignatureListHeaderSize` is true on the first iteration (`0 < 1`), so the code executes:

```go
signature.SignatureData = make([]byte, 0xFFFFFFFF - 16)  // ≈ 4 GiB
```

The process is killed by the OOM killer or panics before `binary.Read` can return an error.

### Fix

After deriving `remainingListSize = SignatureListSize - efiSignatureListHeaderSize`, reject any `SignatureSize` that exceeds it. This bounds every per-entry allocation to bytes actually present in the list, consistent with the UEFI spec (each entry must fit within the list).

### PoC

`TestParseEfiSignatureListOversizedSignatureSize` (added in `events_test.go`) reproduces the issue.

## Test plan

- [x] `go test ./attest/internal/...` passes
- [x] New test `TestParseEfiSignatureListOversizedSignatureSize` verifies the fix

### DIFF
--- a/attest/internal/events_test.go
+++ b/attest/internal/events_test.go
@@ -7,6 +7,45 @@ import (
 	"github.com/google/go-cmp/cmp"
 )
 
+// buildEFISignatureListData builds a raw EFI_SIGNATURE_LIST byte slice for testing.
+// The returned bytes begin with the 28-byte header followed by `payloadLen` zero bytes.
+func buildEFISignatureListData(sigType [16]byte, signatureListSize, signatureHeaderSize, signatureSize uint32, payloadLen int) []byte {
+	buf := &bytes.Buffer{}
+	buf.Write(sigType[:])
+	writeUint32 := func(v uint32) {
+		b := make([]byte, 4)
+		b[0] = byte(v)
+		b[1] = byte(v >> 8)
+		b[2] = byte(v >> 16)
+		b[3] = byte(v >> 24)
+		buf.Write(b)
+	}
+	writeUint32(signatureListSize)
+	writeUint32(signatureHeaderSize)
+	writeUint32(signatureSize)
+	buf.Write(make([]byte, payloadLen))
+	return buf.Bytes()
+}
+
+// TestParseEfiSignatureListOversizedSignatureSize verifies that a crafted
+// EFI_SIGNATURE_LIST with SignatureSize larger than the remaining list space
+// is rejected without attempting a multi-gigabyte allocation.
+// Prior to the fix, setting SignatureSize=0xFFFFFFFF caused a ~4 GiB
+// make([]byte, ...) that crashed or OOM-killed the verifier process.
+func TestParseEfiSignatureListOversizedSignatureSize(t *testing.T) {
+	// certX509SigGUID: {a5c059a1-94e4-4aa7-87b5-ab155c2bf072}
+	certX509SigGUID := [16]byte{0xa1, 0x59, 0xc0, 0xa5, 0xe4, 0x94, 0xa7, 0x4a, 0x87, 0xb5, 0xab, 0x15, 0x5c, 0x2b, 0xf0, 0x72}
+
+	// SignatureListSize=29 (header=28 + 1 payload byte), SignatureSize=0xFFFFFFFF
+	// remainingListSize = 29-28 = 1; SignatureSize (0xFFFFFFFF) > 1 → must error.
+	data := buildEFISignatureListData(certX509SigGUID, 29, 0, 0xFFFFFFFF, 1)
+
+	_, _, err := parseEfiSignatureList(data)
+	if err == nil {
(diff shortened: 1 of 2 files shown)


jlory:
Looks similar to: https://github.com/google/go-attestation/pull/490

attest/internal/events_test.go
```diff
@@ -7,6 +7,45 @@ import (
...
+// is rejected without attempting a multi-gigabyte allocation.
+// Prior to the fix, setting SignatureSize=0xFFFFFFFF caused a ~4 GiB
+// make([]byte, ...) that crashed or OOM-killed the verifier process.
+func TestParseEfiSignatureListOversizedSignatureSize(t *testing.T) {
+	// certX509SigGUID: {a5c059a1-94e4-4aa7-87b5-ab155c2bf072}
+	certX509SigGUID := [16]byte{0xa1, 0x59, 0xc0, 0xa5, 0xe4, 0x94, 0xa7, 0x4a, 0x87, 0xb5, 0xab, 0x15, 0x5c, 0x2b, 0xf0, 0x72}
+
+	// SignatureListSize=29 (header=28 + 1 payload byte), SignatureSize=0xFFFFFFFF
+	// remainingListSize = 29-28 = 1; SignatureSize (0xFFFFFFFF) > 1 → must error.
+	data := buildEFISignatureListData(certX509SigGUID, 29, 0, 0xFFFFFFFF, 1)
```
liamjm: Why set the sig size to 0xFFFFFFFF ? This is unnecessarily large. A value of 2 is sufficient to exercise the new code to prove that it is working.

attest/internal/events_test.go
```diff
@@ -7,6 +7,45 @@ import (
...
+// TestParseEfiSignatureListOversizedSignatureSize verifies that a crafted
+// EFI_SIGNATURE_LIST with SignatureSize larger than the remaining list space
+// is rejected without attempting a multi-gigabyte allocation.
+// Prior to the fix, setting SignatureSize=0xFFFFFFFF caused a ~4 GiB
+// make([]byte, ...) that crashed or OOM-killed the verifier process.
+func TestParseEfiSignatureListOversizedSignatureSize(t *testing.T) {
+	// certX509SigGUID: {a5c059a1-94e4-4aa7-87b5-ab155c2bf072}
+	certX509SigGUID := [16]byte{0xa1, 0x59, 0xc0, 0xa5, 0xe4, 0x94, 0xa7, 0x4a, 0x87, 0xb5, 0xab, 0x15, 0x5c, 0x2b, 0xf0, 0x72}
+
+	// SignatureListSize=29 (header=28 + 1 payload byte), SignatureSize=0xFFFFFFFF
```
liamjm: nit: instead of comments, variables could be used for this, which would allow extra checks and reduce the need to comment., eg

payloadSize := 1
dataSize:= efiSignatureListHeaderSize + payloadSize
// Error, specific a sig Size greater than the payload
sizeSig := payloadSize + 1
data := buildEFISignatureListData(certX509SigGUID, dataSize , 0, payloadSize + 1, payloadSide)
if got := len(data) , got != dataSize {
  t.Fatalf("sig list data is %d bytes but want %d", got, dataSize)
}

What do you think?

liamjm (changes requested):

attest/internal/events_test.go
evilgensec: Good point — updated to use 2 instead.

attest/internal/events_test.go
evilgensec: Makes sense, done. Used named variables (payloadSize, dataSize, sigSize) and added a length check on the built data.

liamjm (approved):

PR #415, opened 2025-03-25
Support ECDSA based AK's on Windows

skepth:
Currently, ECDSA AK support was partially added in the following commits for Linux platform:
- https://github.com/google/go-attestation/commit/1b202b12e80969eb8c5933a9197c16f5f7529d37
- https://github.com/google/go-attestation/commit/c7aee80c5d760a084adf51055b748d04a3187818

This PR builds on the above commits and adds the following functionality:
1. Ability to generate encrypted credential blob for activating ECDSA based AK's.
2. Support minting P256, P384 & P521 curve based ECDSA AK's on Windows platform.
3. Add ability to use custom names for AK's on Windows platform.

PR #504, opened 2026-05-27
Add Windows application key support

hslatman:
This PR adds support for Windows application keys.

It's heavily based on https://github.com/google/go-attestation/pull/274 with the changes from @mjg59 and @brandonweeks, and another attempt at getting (some of my changes from) https://github.com/google/go-attestation/pull/317 in, but made up-to-date with the latest.

New functionalities on Windows:

- Create a new (named) key, including `QualifyingData`. By default, a random name with `app-` prefix is generated.
- Certifying the key using either RSA or EC AKs (EC AK support was added in https://github.com/google/go-attestation/pull/415; didn't exist at the time of the previous PRs)
- Make `crypto.Signer` interface work for RSA and ECDSA on Windows

I have a few more changes lined up, but they'll go in separate PRs.


liamjm:
Thanks!

I have implemented something similar internally, but haven't been able to push it here.

I'll have a look, but it might take a couple of weeks.

hslatman:
Great, thank you @liamjm! 

I'll probably line up the other PRs on top of this branch in the meantime. Will also do some more internal testing to verify porting over the changes result in the same behavior as our current fork.

attest/wrapped_tpm20.go
```diff
@@ -838,6 +839,38 @@ func signECDSA(rw io.ReadWriter, key tpmutil.Handle, digest []byte, curve ellipt
...
+		// This is prevented by selecting an appropriate signature
+		// scheme based on the curve.
+		var h tpm2.Algorithm
+		switch curve {
+		case elliptic.P384():
+			h = tpm2.AlgSHA384
+		case elliptic.P521():
+			h = tpm2.AlgSHA512
+		default:
+			h = tpm2.AlgSHA256
```
liamjm: Can we be explicit here about having a case for curve P256?
Reasons:
a) P224 is a curve that is not covered in the switch
b) future proofing for other cases

The default can return an error of being unsupported.

attest/wrapped_tpm20.go
```diff
@@ -853,7 +886,7 @@ func signECDSA(rw io.ReadWriter, key tpmutil.Handle, digest []byte, curve ellipt
 	// that may have been dropped when converting the digest to an integer
 	digest = ret.FillBytes(digest)
 
-	sig, err := tpm2.Sign(rw, key, "", digest, validation, nil)
+	sig, err := tpm2.Sign(rw, key, "", digest, validation, scheme)
```
liamjm, line 981: can you explain the impact of changing the scheme from non-nil?  How did this work before with a nil scheme?

attest/pcp_windows.go
```diff
@@ -296,63 +300,209 @@ func getPCPCerts(hProv uintptr, propertyName string) ([][]byte, error) {
...
+
+	pubBlob, privBlob, err := decodeKeyBlob(keyBlob)
+	if err != nil {
+		return 0, nil, nil, fmt.Errorf("decodeKeyBlob failed: %v", err)
+	}
+
+	return kh, pubBlob, privBlob, nil
+}
+
+func decodeKeyBlob(keyBlob []byte) ([]byte, []byte, error) {
```
liamjm: a) please document what format is being decoded here.
b) why read each field separately? I find it easier to read if the blob is read in chunks. 

For example:

// PcpKeyBlob Header (Little Endian)
type PcpKeyBlobHeader struct {
	Magic    uint32 // 0x4D504350 ("PCMP")
	CbHeader uint32 // Length of this header (usually 20 or 24 bytes)
	PcpType  uint32 // TPM Version (2 = TPM 2.0)
	Flags    uint32 
	CbTpmKey uint32 // Total byte size of the trailing TPM structures
}

func ExtractTpm20Blobs(data []byte) (pubBlob []byte, privBlob []byte, err error) {
	if len(data) < 20 {
		return nil, nil, errors.New("invalid data: buffer size too small for PCP header")
	}

	var header PcpKeyBlobHeader
	reader := bytes.NewReader(data)

	// 1. Read outer PCP Header (Windows Native - Little Endian)
	if err = binary.Read(reader, binary.LittleEndian, &header); err != nil {
		return nil, nil, fmt.Errorf("failed to read PCP header: %w", err)
	}

	// Structural safety checks
	if header.Magic != BcryptPcpKeyMagic {
		return nil, nil, fmt.Errorf("invalid magic signature: expected 0x%08X, got 0x%08X", BcryptPcpKeyMagic, header.Magic)
	}
	if header.PcpType != PcpTypeTpm20 {
		return nil, nil, errors.New("unsupported hardware type: payload is not a TPM 2.0 key")
	}
....
//etc
}

attest/pcp_windows.go
```diff
@@ -296,63 +300,209 @@ func getPCPCerts(hProv uintptr, propertyName string) ([][]byte, error) {
 	return out, nil
 }
 
-// NewAK creates a persistent attestation key of the specified name.
-func (h *winPCP) NewAK(name string, alg Algorithm) (uintptr, error) {
+func (h *winPCP) newKey(name string, alg string, length uint32, policy uint32) (uintptr, []byte, []byte, error) {
```
liamjm: please document what the return values are, especially the 2 []byte.

attest/pcp_windows.go
```diff
@@ -296,63 +300,209 @@ func getPCPCerts(hProv uintptr, propertyName string) ([][]byte, error) {
...
+	if err := binary.Read(r, binary.BigEndian, &pubKey); err != nil {
+		return nil, nil, fmt.Errorf("failed to read public key: %v", err)
+	}
+
+	privBlob := make([]byte, privLen)
+	if err := binary.Read(r, binary.BigEndian, &privBlob); err != nil {
+		return nil, nil, fmt.Errorf("failed to read private blob: %v", err)
+	}
+
+	return pubKey[2:], privBlob[2:], nil
```
liamjm: nit: comment why the first 2 bytes are being skipped over and/or define this as a constant.

attest/pcp_windows.go
```diff
@@ -296,63 +300,209 @@ func getPCPCerts(hProv uintptr, propertyName string) ([][]byte, error) {
...
+}
+
+// NewAK creates a persistent attestation key of the specified name.
+func (h *winPCP) NewAK(name string, alg Algorithm) (uintptr, error) {
+	kh, _, _, err := h.newKey(name, string(alg), uint32(alg.Size()), nCryptPropertyPCPKeyUsagePolicyIdentity)
+	return kh, err
+}
+
+// NewKey creates a persistent application key of the specified name.
+func (h *winPCP) NewKey(name string, alg Algorithm, size int) (uintptr, []byte, []byte, error) {
```
liamjm, line 399: Why is there more parsing of the alg and size here, but not for NewAK()?

attest/tpm_windows.go
```diff
@@ -248,23 +249,105 @@ func (t *windowsTPM) loadAK(opaqueBlob []byte) (*AK, error) {
...
+
+	// Certify application key by AK
+	certifyOpts := CertifyOpts{QualifyingData: opts.QualifyingData}
+	cp, err := k.certify(t, hnd, certifyOpts)
+	if err != nil {
+		return nil, fmt.Errorf("ak.Certify() failed: %v", err)
+	}
+
+	if !bytes.Equal(pub, cp.Public) {
+		return nil, fmt.Errorf("certified incorrect key, expected: %v, certified: %v", pub, cp.Public)
```
liamjm: nit: consider `% x` for printing bytes as hex chars not decimal.

liamjm (changes requested):
Really great work.

I've given this PR a thorough review and it looks rock solid. However, I haven't got in front of a Windows PC to test it, that will happen next week.

attest/wrapped_tpm20.go
hslatman: IIRC, on Linux the correct scheme to use would be handled on one of the lower layers. On Windows a `nil` scheme would simply fail for ECDSA signing.

It's been a while, and I may be able to find some notes on this to get to the bottom of this again if needed.

attest/tpm_windows.go
hslatman: https://github.com/google/go-attestation/pull/504/commits/b7f2bf66547ff738edfe2fda80877e4f167a28a0

attest/wrapped_tpm20.go
hslatman: https://github.com/google/go-attestation/pull/504/changes/b7f2bf66547ff738edfe2fda80877e4f167a28a0

attest/pcp_windows.go
hslatman: Great catch. The changes from the original PR predated the most recent changes of adding EC AKs on Windows, and I forgot to port that logic over. I did so in https://github.com/google/go-attestation/pull/504/changes/1df4059176b06ea9e9c6edcc5af2c0dc750a01c1.

I've also updated it with handling the new `P256`, `P384` and `P521` algorithms, so that different curves can be used. The current `ECDSA.Size()` always returns `256` (introduced in https://github.com/google/go-attestation/pull/415), and there was no `size` param in `NewAK`, as there was no need so far. It would require extending `AKConfig` with a `Size` property to make that path work. I presume the idea may have been to deprecate the `ECDSA` algorithm identifier, and use the `PXXX` ones? Or maybe the new identifiers are only there to be passed to Windows.

attest/wrapped_tpm20.go
hslatman: Some additional context of the change in our fork: https://github.com/smallstep/go-attestation/pull/5: https://github.com/google/go-attestation/pull/343/changes/1702994f7d221d2b31e054fa308a5f2b35160116.

In our fork we decided to only do it for Windows, but I believe always doing it is also safe, assuming the scheme to use is derived correctly.

attest/pcp_windows.go
hslatman: Addressed in https://github.com/google/go-attestation/pull/504/changes/a3a1865062c2ccad2eb2eefa0c0c6fec16dc4702

attest/pcp_windows.go
hslatman: Addressed in https://github.com/google/go-attestation/pull/504/changes/a3a1865062c2ccad2eb2eefa0c0c6fec16dc4702

attest/pcp_windows.go
hslatman: Refactored this in https://github.com/google/go-attestation/pull/504/changes/a3a1865062c2ccad2eb2eefa0c0c6fec16dc4702. The function is extracted out of the `windows` build, so that it gets tested in CI.

hslatman:
@liamjm I've rebased the PR, and addressed your comments. 

In https://github.com/google/go-attestation/pull/504/commits/072d1407d129a73f7d6b506609d7be017f082c4b I did some more refactoring and made Windows key usage more robust. It also removed a hardcoded SHA1 default to rely on the contents of the AK for determining the hashing algorithm when certifying the key. These are some behavior changes that may require some scrutiny. Results in a small test program on Windows look OK so far on my side.

PR #516, opened 2026-08-10
Implemented the EK-based key attestation protocol.

### LINKED ISSUE #437: Implement EK-Based Key Attestation with TPM Firmware Version
Protocol draft: https://trustedcomputinggroup.org/wp-content/uploads/EK-Based-Key-Attestation-with-TPM-Firmware-Version-V1-RC1_9July2025.pdf

zhsh:
This protocol allows a CA to issue challenges encrypted to a TPM's decryption Endorsement Key (EK). The client decrypts the challenge using TPM2_Import and TPM2_Load, and then certifies its Attestation Key (AK) using the decrypted HMAC key via TPM2_Certify.

Detailed changes:

- Added DecryptionEkChallenge struct to attest package to hold challenge parameters (ephemeral public key, encrypted sensitive area, duplicate public area).
- Added CertifyWithDecryptionEk to AK struct, which delegates to the internal ak interface.
- Implemented certifyWithDecryptionEk for wrappedKey20 (TPM 2.0 on Linux). It uses a policy session to authorize the EK, imports the challenge, loads the imported key, and calls CertifyEx to sign the AK name.
- Implemented certifyWithDecryptionEk for Windows to return errUnsupported.
- Added a comprehensive simulator test TestSimCertifyWithDecryptionEk that acts as the CA (generating the challenge) and verifies the client's output (validating the HMAC signature and attestation data).


attest/ek_challenge.go
```diff
@@ -0,0 +1,285 @@
...
+)
+
+// HMAC represents a SHA-256 HMAC digest, strictly bound to the SHA-256 output size (32 bytes).
+type HMAC [sha256.Size]byte
+
+func generateRestrictedHMACKey() (tpm2.TPMTPublic, tpm2.TPMTSensitive, HMAC) {
+	var hmacKeyBytes HMAC
+	rand.Read(hmacKeyBytes[:])
+
+	obfuscate := make([]byte, 32)
```
liamjm, line 40: Is 32 special? Is this meant to be the same size as sha256.Size?
We should document or explain this constant.

attest/ek_challenge.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+	if attest.Type != tpm2.TPMSTAttestCertify {
+		return fmt.Errorf("unexpected type %0x", attest.Type)
+	}
+	return nil
+}
+
+func verifyCertify(name *tpm2.TPM2BName, certifyInfo *tpm2.TPMSCertifyInfo) error {
+	// Check that the certified Name is the same as we expected.
+	if !bytes.Equal(name.Buffer, certifyInfo.Name.Buffer) {
```
liamjm: should we use subtle.ConstantTimeCompare for these as well? It is only the name so I guess there are no timing attacks here.

liamjm (commented):
Just some nits.

attest/ek_challenge.go
zhsh: Done.

attest/ek_challenge.go
zhsh: Done.

liamjm (approved):

